PR #41, opened 2017-04-10
Username is now required for make create_user

Ydrasil:
If we don't set username, then we can't create 2 profiles one after another

### DIFF
--- a/user/models.py
+++ b/user/models.py
@@ -43,5 +43,7 @@ class User(UserMixin, db.Document):
         return str(self.username or 'Unknown')
 
     def save(self, *args, **kwargs):
+        if self.username is None:
+            self.username = self.email.split('@')[0]
         self.slug = slugify(self.username)
         return super().save(*args, **kwargs)


vinyll:
That's a good point. I was thinking of creating a default username based on the email: `username = email.split('@')[0]`.
Any other opinion?

Ydrasil:
With this way (`username = email.split('@')[0]`) 2 peoples can have the same username (i.e. jack@email.com and jack@yopmail.com)

I can even prevent someone to join neomad, if I change my username accordingly with his email (niark! niark!)

vinyll:
yes. We should make the username unique and the slug unique as well.
The issue is more obvious with the slug than the username.
We could have 3 different usernames: _john.doe_, _John Doe_, _John Doé_ and the same slug: _john-doe_.

The question of not permitting a used username/slug is more complex as:
- it should be managed from the command `make create_user`
- it should be managed from the frontend (https://neomad.org/@my-profile) and display a user friendly message.

I guess we should:
1. update the `User` model to make `username` unique (and `slug`).
2. catch the error in the `make create_user` command line.
3. display a user-friendly error on the frontend when a user updates to a already taken username.

vinyll:
@Ydrasil: Because it's a lot more work if we want to consider the uniqueness of the username and the slug, why don't we just generate a username from the email with the script you suggest for now?
It will at least solve what you were trying to solve in 99% of cases.

Ydrasil:
But it still 1% of cases that doesn't work

vinyll:
Yes. Right now we have 50% of cases that don't work.
Doing this simple solution we solve 99%.
If we want to solve 99.9%, we'll need much more code and more heavy testing.

It does not mean we won't need to do it, but that we could make it [KISS](https://en.wikipedia.org/wiki/KISS_principle#In_software_development) and improve it later.

Ydrasil:
That's a good point

vinyll:
What should we do with this PR @Ydrasil?
Do you want to continue it or should we close?

Ydrasil:
@vinyll I think that, we can close this PR but add an issue for add an username. 
You have right with the fact that we'll need much more code for add an username and now we need to avoid an issue with the "none" as username.

vinyll:
The task `create_user` is no more needed. We have a _/signup_ url now.

PR #42, opened 2017-04-10
Localize around

fuentesloic:
- Find your current place and put a red marker (in front of all of them)
- Put marker and information about interesting place from `/statics/test.json`
- Center view and zoom according your current position


static/js/around-map.js
```diff
@@ -0,0 +1,46 @@
+mapboxgl.accessToken = 'pk.eyJ1IjoibmVvbWFkIiwiYSI6ImNqMHRrZ3ZwdzAwNDgzMm1kcHRhMDdsZGIifQ.bOSlLkmc1LBv0xAbcZXpog';
+const map = new mapboxgl.Map({
+  container: 'map',
+  style: 'mapbox://styles/mapbox/streets-v9',
+  center: [-10, 45],
+  zoom: 2
+})
+
+// users interesting points
+const giveData = new Worker("/static/js/webworker-around.js");
```
vinyll: I would rename this to `pois`, or `workspots`, or something more related to what it actually is.

static/js/around-map.js
```diff
@@ -0,0 +1,46 @@
...
+
+giveData.onmessage = (informations) => {
+  const [name,wifi,power,comment,lng,lat] = informations.data
```
vinyll: space after commas

static/js/around-map.js
```diff
@@ -0,0 +1,46 @@
...
+
+  const el = document.createElement('div')
+  el.classList.add('marker')
+
+  let content = "<h2>"+name+"</h2><p>Wifi quality: "+ wifi+"<br>Power available: "+power+"<br>Comments: "+comment+"</p>"
```
vinyll: You're using ES6, therefore you should use [template strings](https://babeljs.io/learn-es2015/#ecmascript-2015-features-template-strings)

vinyll: Maybe the wifi , power and comments should be in a list (`<li>`) instead of `<p>`

static/js/around-map.js
```diff
@@ -0,0 +1,46 @@
...
+  const popup = new mapboxgl.Popup({offset: [10, -20]})
+      .setHTML(content)
+
+  new mapboxgl.Marker(el, {offset:[0, -30]})
+      .setLngLat([lng, lat])
+      .setPopup(popup)
+      .addTo(map)
+};
+
+giveData.postMessage("info-requested");
```
vinyll: quote inconsistency here. You should use single quotes like everywhere else.

static/js/around-map.js
```diff
@@ -0,0 +1,46 @@
...
+
+// user last location
+if(current_location.length) {
+  const popup = new mapboxgl.Popup({offset: [10, -20]})
+      .setText('you')
+
+  const el = document.createElement('div');
+  el.classList.add('marker')
+  el.classList.add('you')
```
vinyll: _you_ as a class name is not explicit. I would use something like _current-location_ or _current_ eventually.

static/js/around-map.js
```diff
@@ -0,0 +1,46 @@
...
+      .setPopup(popup)
+      .addTo(map)
+};
+
+giveData.postMessage("info-requested");
+
+// user last location
+if(current_location.length) {
+  const popup = new mapboxgl.Popup({offset: [10, -20]})
+      .setText('you')
```
vinyll: What do you think of a text like "Your current location"?

static/js/around-map.js
```diff
@@ -0,0 +1,46 @@
...
+  new mapboxgl.Marker(el, {offset:[0, -30]})
+      .setLngLat([lng, lat])
+      .setPopup(popup)
+      .addTo(map)
+};
+
+giveData.postMessage("info-requested");
+
+// user last location
+if(current_location.length) {
```
vinyll: I know I made the error and I'll need to fix it, but JS takes camelCase writing, not underscores in its variable names. it should therefore be called `currentLocation` everywhere in JS.
Python does take underscore.

static/js/webworker-around.js
```diff
@@ -0,0 +1,18 @@
+onmessage = _ => {
+  fetch("/static/test.json")
```
vinyll: single quotes

static/js/webworker-around.js
```diff
@@ -0,0 +1,18 @@
...
+    .then(response => {
```
vinyll: shorter version:

```js
then(response => response.json())
```

static/js/webworker-around.js
```diff
@@ -0,0 +1,18 @@
...
+      return response.json();
+    })
+    .then(json => {
+      const items = json;
```
vinyll: We do not use semi-columns ";" in this project unless very necessary.

static/js/around-map.js
```diff
@@ -0,0 +1,46 @@
...
+  container: 'map',
+  style: 'mapbox://styles/mapbox/streets-v9',
+  center: [-10, 45],
+  zoom: 2
+})
+
+// users interesting points
+const giveData = new Worker("/static/js/webworker-around.js");
+
+giveData.onmessage = (informations) => {
```
vinyll: parenthesis are not required here. For consistency – because they are not used elsewhere in the code when it's a unique parameter – they need to be removed.

vinyll (changes requested):
Mostly syntax adaptation.
Meaning the code is good!! 👍

static/js/around-map.js
fuentesloic: I tried to change in the "user" to save the position into a currentLocation but I have all the time an mongoError ... pretty hard to fix it :D
So I don't correct this one, an when you will fix it I push a correct with currentLocation

static/css/around-map.css
```diff
@@ -1,4 +1,4 @@
-.marker.you {
+.marker.current-location {
     background-image: url('/static/images/marker-red.png');
     z-index: 9999;
```
vinyll: Do not make a z-index that is so huge. I understand it should be over everything, but what if everything else like an alert banner needs to be over? And then a popup window over that banner alert?
Whe end up with ridiculous counts of _9_s.
Instead why don't you just apply a z-index on the markers (eg: `z-index: 2`), and 1 higher on the `.current-location`?

static/js/around-map.js
```diff
@@ -7,15 +7,21 @@ const map = new mapboxgl.Map({
...
   el.classList.add('marker')
 
-  let content = "<h2>"+name+"</h2><p>Wifi quality: "+ wifi+"<br>Power available: "+power+"<br>Comments: "+comment+"</p>"
+  let content = `<h2>Hello ${name}</h2>
+                 <ul>
+                    <li>Wifi quality: ${wifi}</li>
+                    <li>Power available: ${power}</li>
+                    <li>Comments: ${comment}</li>
+                  </ul>`
+
```
vinyll, line 24: Looks 👌

static/js/around-map.js
```diff
@@ -25,22 +31,22 @@ giveData.onmessage = (informations) => {
...
 
-giveData.postMessage("info-requested");
+pois.postMessage('info-requested');
 
 // user last location
-if(current_location.length) {
+if(currentLocation.length) {
   const popup = new mapboxgl.Popup({offset: [10, -20]})
-      .setText('you')
+      .setText('You current location')
```
vinyll: _You**r** current location_

static/js/webworker-around.js
```diff
@@ -1,18 +1,16 @@
 onmessage = _ => {
-  fetch("/static/test.json")
-    .then(response => {
-      return response.json();
-    })
+  fetch('/static/test.json')
```
vinyll: Now that you're a big boy, maybe is it time for you to read directly from the API… ;)
`fetch('http://158.58.170.155:3030/spots')`
This way you can get rid of the _test.json_ file 😌

static/js/webworker-around.js
```diff
@@ -1,18 +1,16 @@
...
+    .then(response => response.json())
     .then(json => {
-      const items = json;
+      const items = json
```
vinyll:
```js
.then(items => {
  items.forEach(item => {
    …
```

vinyll (changes requested):
We're getting there! :)

vinyll (changes requested):
It's all perfect, you made it! 🍾 

Now a challenge before merging: 25 commits for one single feature is **a lot**.
You may want to use `git rebase -i a71cf5d9^` and edit which commit you want to squash and which you want to pick. Maybe 2 to 4 commits for the frontend would be enough. (see https://www.atlassian.com/git/tutorials/rewriting-history#git-rebase-i for examples)

fuentesloic:
@vinyll : what do you think now about this "squash" ?

static/js/webworker-around.js
```diff
@@ -0,0 +1,16 @@
...
+    .then(response => response.json())
+    .then(items => {
+      items.forEach(item => {
+        let name = item.name
+        let wifi = item.wifiQuality
+        let power = item.powerAvailable
+        let comment = item.comments[0]
+        let lng = item.position.longitude
+        let lat = item.position.latitude
+        postMessage([name,wifi,power,comment,lng,lat])
```
vinyll: space after commas

vinyll (approved):
Commits look good. You just got it!  🎉 
Just one notice about commas in JS (just `push -f` here not to make a new commit only for that).

And you're good to merge!

static/js/webworker-around.js
fuentesloic: oups I forget ... I change that right now and I'll do a `git push -f`and a merge :) youhouuuu ^^

PR #43, opened 2017-04-10
The system can detect the language used in articles

Ydrasil:


templates/layout.html
```diff
@@ -1,4 +1,5 @@
 <!DOCTYPE html>
+{% block language %}{% endblock %}
```
vinyll:
```
<html lang={% block language %}en{% endblock %}>
```
with a closing `</html>` at the end of the document.

vinyll: The `en` here will be the default language

templates/blog/article.html
```diff
@@ -1,5 +1,7 @@
 {% extends 'layout.html' %}
 
+{% block language %}<html lang="{{ article.language }}">{% endblock %}
```
vinyll:
```
{% block language %}{{ article.language }}{% endblock %}
```

vinyll (changes requested):
While you're there, it would be nice to have:
- links pointing to an article with the `lang={{ article.language }}` attribute
- each `article` element in list of articles (checkout the *article_list.html* template) with a `lang={{ article.language }}` attribute.

---

This code is easy to read and review. Seems soooo clean! 💃

vinyll (approved):
That's a cool feature!!!
🥇

PR #60, opened 2017-04-15
add button to re-center based on current user position

fuentesloic:
that works according if the user allowed neomad to track him or not


static/css/around-map.css
```diff
@@ -6,3 +6,16 @@
 .mapboxgl-popup {
     z-index: 2;
 }
+
+button {
```
vinyll: Bewarn that this selector is so generic that it will inevitability interfere with other buttons on the page.

templates/around/map.html
```diff
@@ -11,6 +11,9 @@
 
 {% block main %}
   <h1>Find someone or something close to <em>you</em></h1>
+  <button type="button" onclick="userCenter()">
```
vinyll: remove the quotes

templates/around/map.html
```diff
@@ -11,6 +11,9 @@
...
+    <img height="100%" src="/static/images/recenter.png"/>
```
vinyll: remove the quotes and the ending slash ;)
(to be consistent with the writing of the code in Neomad)

templates/around/map.html
vinyll: `<buton type=button`?

static/js/around-map.js
```diff
@@ -43,10 +43,22 @@ if(currentLocation.length) {
...
+    .setLngLat(position)
+    .setPopup(popup)
+    .addTo(map)
 
-  map.setCenter(currentLocation)
+  map.setCenter(position)
   map.setZoom(11)
 }
+
+userCenter = _  => {
```
vinyll: `const userCenter =…`

static/js/around-map.js
```diff
@@ -43,10 +43,22 @@ if(currentLocation.length) {
...
+  if(currentLocation.length > 0) {
+    currentMarker(currentLocation.reverse())
+  } else {
+    navigator.geolocation.getCurrentPosition(function(position) {
```
vinyll: remove "function"

static/js/around-map.js
```diff
@@ -43,10 +43,22 @@ if(currentLocation.length) {
...
+      currentMarker([position.coords.longitude, position.coords.latitude]);
```
vinyll: remove semi-column

static/js/around-map.js
```diff
@@ -43,10 +43,22 @@ if(currentLocation.length) {
...
+    });
```
vinyll: remove semi-column

static/css/around-map.css
```diff
@@ -6,3 +6,16 @@
...
+    height: 50px;
+    width: 50px;
+    position: absolute;
+    right: 1em;
```
vinyll: use `rem` instead (@newick's recommendation)

static/css/around-map.css
```diff
@@ -6,3 +6,16 @@
...
+    margin-top: 1em;
```
vinyll: use `rem` instead

static/css/around-map.css
```diff
@@ -6,3 +6,16 @@
 .mapboxgl-popup {
     z-index: 2;
 }
+
+button {
+    height: 50px;
+    width: 50px;
+    position: absolute;
```
vinyll: Does this work well? What is it absolute to? (what is the relative parent?)

vinyll (changes requested):
A few syntax corrections, a question and some css selector to be more precise.
Meaning it looks great :)

templates/around/map.html
```diff
@@ -11,6 +11,9 @@
 
 {% block main %}
   <h1>Find someone or something close to <em>you</em></h1>
+  <button id=userCenter onclick=userCenter()>
```
vinyll: you're gonna love it: in html and css, the convention is to use dash (-) or underscode (_).
In this project we use dash.
Therefore the id should be _user-center_ and same in the css ;)

vinyll (changes requested):
very close!

templates/around/map.html
fuentesloic: Oh I see ! thx

vinyll (approved):
🎉 🌮

vinyll:
Well, it's approved but you need to squash your commits @fuentesloic  ;)

PR #62, opened 2017-04-16
fixed social profile infos editing and displaying

vinyll:
## Editor mode
screen shot 2017-04-16 at 9 18 31 pm

---

## Visitor mode
screen shot 2017-04-16 at 9 19 03 pm


newick (approved):
ok for me

PR #63, opened 2017-04-17
focalize in the user when you reach /around

fuentesloic:
- generate an hash with the ID point when you click on share
- if hash in url, find wich point is it and move on it


vinyll:
I would suggest this icon instead: https://commons.wikimedia.org/wiki/File:Ic_my_location_48px.svg with a white background (à-la gMaps).
200px-ic_my_location_48px svg

vinyll:
I'm localized in Tanzania. Seems like _lat_ and _lng_ have been flipped.

static/js/around-map.js
```diff
@@ -7,34 +7,33 @@ const map = new mapboxgl.Map({
...
-    .addTo(map)
-};
+    new mapboxgl.Marker(el, {offset:[0, -30]})
+      .setLngLat([poi.position.longitude, poi.position.latitude])
+      .setPopup(popup)
+      .addTo(map)
+  })
+}
 
 pois.postMessage('info-requested')
```
fuentesloic: if you create `const worker = new Worker('...')`, you should use `worker.postMessage('...')` isn't ? 

I changed it ;) that was to do a review to you ! mouhahahahahahahahah

vinyll:
Screenshot?

fuentesloic:
capture d ecran 2017-04-19 a 09 23 33

vinyll (commented):
I would have used _function_ to declare functions. eg: `function userCenter () {`. I feel like it's more explicit and readable that using a _const_. No big deal for now though.

Squashing the commits will be essential though for not having 11.
We'll need to consider https://www.mapbox.com/mapbox-gl-js/example/locate-user/ in a next PR

static/js/around-map.js
```diff
@@ -6,6 +6,8 @@ const map = new mapboxgl.Map({
   zoom: 2
 })
 
+map.addControl(new mapboxgl.NavigationControl());
```
vinyll: no semi-column ;)

vinyll (changes requested):

static/js/around-map.js
fuentesloic: done ... still bad habits

static/js/around-map.js
```diff
@@ -7,6 +7,7 @@ const map = new mapboxgl.Map({
 })
 
 map.addControl(new mapboxgl.NavigationControl());
+map.addControl(new mapboxgl.GeolocateControl());
```
vinyll: Again, no semi-column ;)

static/js/around-map.js
```diff
@@ -22,7 +23,7 @@ worker.onmessage = response => {
                       <li>Wifi quality: ${poi.wifiQuality}</li>
                       <li>Power available: ${poi.powerAvailable}</li>
                     </ul>`
-              )
+            )
```
vinyll: What is this aligned to?

vinyll (changes requested):

PR #69, opened 2017-04-19
/around - navigation control, hash url and refactorisation

fuentesloic:


fuentesloic:
and now ? I hope it ok ^^

static/js/around-map.js
```diff
@@ -61,4 +62,27 @@ const userCenter = _  => {
...
 }
 
-userCenter()
+const sharePosition = id => {
+  window.location.hash = id
+}
+
+// center the map according context
+if (window.location.hash.indexOf('#') == 0) {
+  const hash = window.location.hash.split('#')[1]
```
vinyll: `const hash = window.location.hash.slice(1)` seems more readable

static/js/around-map.js
```diff
@@ -61,4 +62,27 @@ const userCenter = _  => {
...
+  fetch('/around/spots.json')
+    .then(response => response.json())
+    .then(items => {
+      const hashData = items.find(item => item._id == hash)
+      const hashLng = hashData.position.longitude
+      const hashLat = hashData.position.latitude
+
+      const lng = parseFloat(hashLng)
+      const lat = parseFloat(hashLat)
```
vinyll: you could just write shorter:
```
const lng = parseFloat(hashData.position.longitude)
const lat = parseFloat(hashData.position.latitude)
```	
We same 2 lines, 2 variables and preserve readability :)

static/js/around-map.js
```diff
@@ -61,4 +62,27 @@ const userCenter = _  => {
...
+
+      const hashCoords = [lng, lat]
+      map.setCenter(hashCoords)
```
vinyll: `map.setCenter([lng, lat])` would get rid of the line before

static/js/around-map.js
```diff
@@ -61,4 +62,27 @@ const userCenter = _  => {
...
+      map.setZoom(16)
```
vinyll: I believed that's too zoomed. 13 seemed like a better compromise

vinyll:
The comment says "/around - add navigation control ", but the code does much more than that, it's also a refactoring of the previous code, and implement the hash functionality.

vinyll (changes requested):
Let's just simplify a little the code with the comments given.
It's very not much and then let's just merge this.

static/js/around-map.js
```diff
@@ -6,39 +6,40 @@ const map = new mapboxgl.Map({
...
+                      <li>Power available: ${poi.powerAvailable}</li>
+                    </ul>`
+            )
 
-  new mapboxgl.Marker(el, {offset:[0, -30]})
-    .setLngLat([lng, lat])
-    .setPopup(popup)
-    .addTo(map)
-};
+    new mapboxgl.Marker(el, {offset:[4, -6})
```
vinyll: `new mapboxgl.Marker(el, {offset: [4, -6]})`
A space after the comma and the closing bracket is missing

static/js/around-map.js
```diff
@@ -6,39 +6,40 @@ const map = new mapboxgl.Map({
...
 
-  const popup = new mapboxgl.Popup({offset: [10, -20]})
-    .setHTML(content)
+    const popup = new mapboxgl.Popup({offset: [10, -20]})
+      .setHTML(`<h2>Hello ${poi.name}</h2>
+                   <ul>
+                      <li>Wifi quality: ${poi.wifiQuality}</li>
+                      <li>Power available: ${poi.powerAvailable}</li>
+                    </ul>`
+            )
```
vinyll: What is this aligned to?

static/js/around-map.js
```diff
@@ -6,39 +6,40 @@ const map = new mapboxgl.Map({
...
-
-pois.onmessage = informations => {
-  const [name, wifi, power, comment, lng, lat] = informations.data
+map.addControl(new mapboxgl.NavigationControl())
+map.addControl(new mapboxgl.GeolocateControl())
 
-  const el = document.createElement('div')
-  el.classList.add('marker')
+// users interesting points
+const worker = new Worker('/static/js/webworker-around.js')
```
vinyll, line 13: Use the api, not some static file

vinyll (changes requested):
The URL to the json file is wrong and the JS code was broken.
No chance to run in a browser ;)

static/js/around-map.js
fuentesloic: the worker grab information from the API

static/js/around-map.js
fuentesloic: regression ...

static/js/around-map.js
vinyll: right. I wonder if you could not use a relative import here?
`const worker = new Worker('./webworker-around.js')`

vinyll (changes requested):
I all looks fine on a code perspective.
The only thing that is missing is the user pin cursor.
Right now we have a dot for the user just like the other spots

static/js/around-map.js
fuentesloic: Oh I see now that you mean but it's not working:

> NetworkError: Failed to load worker script at "./webworker-around.js"

`I tried a lot of possibility ... any worked

below the result with `const worker = new Worker('./webworker-around.js')`

fuentesloic:
> Right now we have a dot for the user just like the other spots

I have a special CSS for around which overwrite yours (localize on trips.css)
So I still keep my cursor to the current user position for exemple.
I tryed to uniformize our two css into one and it's possible but maybe have juste one css file for both (like map.css) is a problem for you ?

static/js/around-map.js
fuentesloic: capture d ecran 2017-04-21 a 08 58 04

vinyll: Well, that seems to answer that you cannot use relative imports in WebWorkers ;)

vinyll (approved):
Seems good with me.
We will need to merge the 2 CSS files afterward.

PR #71, opened 2017-04-22
merge css

fuentesloic:


vinyll (approved):
I didn't test if it broke the rendering but I'll assume you have looked everything is still rendering fine.

I guess you could also take #79 and #78 in a next step so that you'll have completed a clean map 👍

fuentesloic:
for now I will destruct this PR because I'm wondering about to have only one CSS ... I'm not sure it's a good idea or not

vinyll:
I believe we should have a CSS that is shared at least for common styles (like the markers, popups and some other common items).
Eventually we'll end with some specific styles for each page that would go elsewhere.

fuentesloic:
yes I wondering exactly the same !! So I open an issue for that I will do that after 'cause it's not really important now.
After features `around-rooterURL`  and `around-appPois` I will work on it

PR #73, opened 2017-04-24
Add circleci file for neomad

Ydrasil:


circle.yml
```diff
@@ -0,0 +1,11 @@
+machine:
+  python:
+    version: 3.5.0
```
vinyll, line 3: Can't you just tell 3.5 instead of 3.5.0 (we want it to work with the latest version).
Eventually also 3.6

vinyll:
ok, let's merge this and improve later…

circle.yml
Ydrasil: So we have to change this line into `version: 3` ?

vinyll:
We won't need it for now, we're on py3.5 all the way 👍

PR #80, opened 2017-04-27
Around hash

fuentesloic:


fuentesloic:
# It's not a real PR !!! 
I want just have a comparaison between master and here

PR #84, opened 2017-04-30
Remove folder's pictures when deleting an article

Ydrasil:


blog/models.py
```diff
@@ -81,6 +82,11 @@ def image(self):
         if len(self.images):
             return self.images[0]
 
+    def delete(self):
```
vinyll: Because you are overriding the `delete` method that is native, you're not sure what it arguments are passing. You should therefore relay them:
```def delete(self, *args, **kwargs)```

You're looping into a recursion when calling `delete()` in the `delete()` method.
When overriding a method you should anyways call the parent method that does a bunch of things:

```
super(Article, self).delete(*args, **kwargs)
```

**Review by vinyll — CHANGES_REQUESTED** [review]
We need a test to cover this as it may get dangerous deleting files.

File: blog/models.py (hunk 1 above)
Comment (Ydrasil) [reply blog/models.py]: I've try the code before and after the changes and the code works good. The article is correctly remove and the picture's folders too

Comment (vinyll) [reply blog/models.py]: I'm sure you have manually tested and it works fine today.
Adding a test [in the articles test file](https://github.com/neomad-team/neomad.org/blob/master/blog/test.py) would permit:
- to avoid any regression when we will update the model and avoid manually test again in the future when updating the code
- to code-prove users that we *actually* delete their files when they request us to do so
- to get CircleCI performing the test of this sensitive and invisible feature for us
- to ensure we don't ever saturate the server with orphan files.

We may do this in an additional commit if it feels better for you.
I can also pair on writing test for this feature or take it over.
Let me know ;)

File: blog/models.py [hunk 2]
```diff
@@ -81,6 +82,11 @@ def image(self):
         if len(self.images):
             return self.images[0]
 
+    def delete(self, *args, **kwargs):
+        super(Article, self).delete(*args, **kwargs)
```
vinyll: When overriding a method and still calling its parent, we should always at least pass the same parameters (that you now do) but also return what the parent returns. It may be `None` today but a boolean, a count or other values in the future parent method updates.
Example of how to achieve this:
```
parent = super(…
…
return parent
```

vinyll (changes requested):

blog/models.py
Ydrasil: I'll try on my side, if I get stuck on something too long, I'll say it to you :wink:

Ydrasil:
@vinyll Now I think its ok :pray:

vinyll:
I guess you forgot to commit @Ydrasil, there are 3 new empty files.
Why are you creating /test/ folder instead of using the already running and dedicated file at https://github.com/neomad-team/neomad.org/blob/master/blog/test.py?

you should be able to run tests with `python -m unittest`

Ydrasil:
My bad, its commited now

test/blog/articles.py
```diff
@@ -0,0 +1,20 @@
+import os
+import unittest
+
+from bin.fixtures import create_article
+from blog.models import Article
+
+
+class ArticleTest(unittest.TestCase):
```
vinyll: This class [already exists](https://github.com/neomad-team/neomad.org/blob/master/blog/test.py#L6)
You should append your method to that class.

blog/test.py
```diff
@@ -1,6 +1,8 @@
+import os
 from unittest import TestCase
 
 from .models import Article
+from bin.fixtures import create_article
```
vinyll: A lib or test should not call a bin script.
You could simply create a _dummy_ article. See below.

blog/test.py
```diff
@@ -20,3 +22,16 @@ def test_content_cleaned_html(self):
         article = Article(title='',
                           content='<p><em>emphased content</em><i>also emphased</i></p>').save()
         self.assertEqual(article.content, '<p><em>emphased content</em><em>also emphased</em></p>')
+
+    def test_delete_article_and_folders_pictures(self):
```
vinyll: Just make it simple:

1. create an article [like we do here](https://github.com/neomad-team/neomad.org/blob/master/blog/test.py#L8)
2. check the folder exists
3. delete the article
4. ensure the folder is deleted

All in all, it would look like:
```
article = Article(title='', content='<p>content</p>').save()
self.assertTrue(os.path.isdir(article.get_images_path())
article.delete()
self.assertFalse(os.path.isdir(article.get_images_path())
```

vinyll (changes requested):
- Changing the _bin/fixtures.py_ might be interesting for duplication but it out of the scope of this PR (deleting an article).
- Do not import a bin file into a lib nor test. bin files are final execution files and subject to change or deletion.
- KISS your test ;)

vinyll:
_bin_ file should be reset and not imported into a module.

Ydrasil:
@vinyll Changes done and conflict resolved

vinyll:
Do you still need to `import os` in the test file?
I guess in this PR you could also reset the bin file to its original version?

Ydrasil:
I use `os` in a test, so I think that I can't remove the `import os`
I'll reset the bin file in this PR

vinyll (approved):

PR #93, opened 2017-05-10
Fix bug from /articles when there is 0 articles

Ydrasil:


Ydrasil:
I haven't found another way to avoid this issue without duplicate code (like if, description or title block)

templates/blog/article_list.html
```diff
@@ -1,15 +1,23 @@
 {% from 'blog/macros.html' import article_list with context %}
 {% extends 'layout.html' %}
 
-{% block title %}Nomad articles - {{ articles[0].title|truncate(40) }}{% endblock %}
+  {% block title %}Nomad articles
```
vinyll: I guess you can wrap all the metas within `{% if articles[0] %}` `{% endif %}`.
If there's no article, there isn't much point of having them anyways.

vinyll (changes requested):

templates/blog/article_list.html
Ydrasil: The things is that I don't succeed to put an if around a block for not displaying it 
For instance 
```
{% block extra_og %}
    {% if articles[0] is defined %}
        <meta property=og:image content="{{ articles[0].image }}">
        <meta property=og:title content="Latest articles ({{ articles[0].title }})">
        <meta property=og:description content="{{ articles[0]|striptags|truncate(200) }}">
    {% endif %}
{% endblock %}
```
If I put the `if` around the `block extra_og` I have a bug. And its the same with other blocks.
I haven't find a way to surround all the metas with only one `{% if articles[0] %} {% endif %}`

And do you prefer `{% if articles[0] %} {% endif %}` or `{% if articles[0] is defined %} {% endif %}`?

vinyll: `{% if articles[0] %}` is more common.

vinyll: It may be that we can't get a _block_ inside a _if_. Good to know.

vinyll: Here you could write:

    {% block title %}Nomad articles {% if articles[0] %} - {{ articles[0].title|truncate(40) }}{% endif %}

on one line

Ydrasil: Done :wink:

templates/blog/article_list.html
```diff
@@ -1,14 +1,19 @@
...
-
-{% block extra_og %}
-<meta property=og:image content="{{ articles[0].image }}">
-<meta property=og:title content="Latest articles ({{ articles[0].title }})">
-<meta property=og:description content="{{ articles[0]|striptags|truncate(200) }}">
+{% block title %}Nomad articles {% if articles[0] %} - {{ articles[0].title|truncate(40) }}{% endif %}
+{% endblock %}
+{% block description %}Read the latest Nomad's articles on Neomad.
+  {% if articles[0] %}
+    Hotest article is « {{ articles[0].title }} ».
```
vinyll: `{% if articles[0] %}Hotest article is « {{ articles[0].title }} ».{% endif %}`

vinyll (changes requested):

templates/blog/article_list.html
```diff
@@ -1,14 +1,17 @@
...
-{% block description %}Read the latest Nomad's articles on Neomad. Hotest article is « {{ articles[0].title }} ».{% endblock %}
-
-{% block extra_og %}
-<meta property=og:image content="{{ articles[0].image }}">
-<meta property=og:title content="Latest articles ({{ articles[0].title }})">
-<meta property=og:description content="{{ articles[0]|striptags|truncate(200) }}">
+{% block title %}Nomad articles {% if articles[0] %} - {{ articles[0].title|truncate(40) }}{% endif %}
+{% endblock %}
+{% block description %}Read the latest Nomad's articles on Neomad.
+  {% if articles[0] %} Hotest article is « {{ articles[0].title }} ».{% endif %}
```
vinyll: the `{%.endblock %}` for the block description should be here

templates/blog/article_list.html
```diff
@@ -1,14 +1,17 @@
...
+    {% block extra_og %}
+      {% if articles[0] %}
+      <meta property=og:image content="{{ articles[0].image }}">
+      <meta property=og:title content="Latest articles ({{ articles[0].title }})">
+      <meta property=og:description content="{{ articles[0]|striptags|truncate(200) }}">
+      {% endif %}
+    {% endblock %}
 {% endblock %}
```
vinyll: the block extra_og should not be embedded into the block description.

vinyll (changes requested):

Ydrasil:
@vinyll My bad about the embedded `block extra_og`, its good now :+1:

vinyll (approved):

PR #95, opened 2017-05-11
Add privacy page and trip deletion

Ydrasil:
I don't know if its a good idea to put the privacy's link page at the left of the logout button.


templates/user/privacy.html
```diff
@@ -0,0 +1,49 @@
...
+    <th>Delete</th>
+  </tr>
+{% for location in locations %}
+  <tr>
+    <td>{{ location.date }}</td>
+    <td>{{ location.position }}</td>
+    <td><form method=post action={{ url_for('privacy_delete_trip',
+                                            date=location.date,
+                                            location=location.position) }}>
+      <button class="button delete">Delete</button></form></td>
```
vinyll: `<button class=button…>`?

user/views.py
```diff
@@ -1,3 +1,5 @@
+from json import loads
```
vinyll: just use `import json`. a method `loads` is a too generic name, we always import json and call function onto the module.

user/views.py
```diff
@@ -18,6 +20,29 @@ def profile(username):
...
+def privacy():
+    user = User.objects.get(id=current_user.id)
+    return render_template('user/privacy.html',
+                           user=user,
+                           locations=user.locations)
+
+
+@app.route('/privacy/<string:location>/<string:date>', methods=['post'])
+@login_required
+def privacy_delete_trip(location, date):
```
vinyll: Did you hand test it? Does it work?

vinyll (changes requested):
Please change the `import json`.
We could approve the PR then.
However we do really need written tests to prove users we actually delete the data.
If you don't feel like taking the written test I can make it. It may be another commit anyways (it could even be a commit before this one, and this one fixing the tests).

user/views.py
Ydrasil: I have hand test it, and its working, but I'll made some test

vinyll: That would be awesome for sure!
For the rest it's all good with me 👍

templates/user/privacy.html
```diff
@@ -0,0 +1,49 @@
...
+  </tr>
+{% endfor %}
+</table>
+
+<script>
+  var current_location = {{ user.current_location or [] }}
+  var locations = [
+  {% for location in locations %}
+    {
+      date: '{{ location.date|datetime }}',
```
vinyll: You may want to insert this into a JS `new Date()` and format it

trips/test.py
```diff
@@ -0,0 +1,66 @@
...
+    def test_trip_exactly_same_area(self):
+        self.client.post('/trips/add', data=json.dumps(self.lat_lng),
+                         content_type='application/json')
+        result = self.client.post('/trips/add', data=json.dumps(self.lat_lng),
+                                  content_type='application/json')
+        user = User.objects.first()
+        self.assertEqual(user.locations.count(), 1)
+        self.assertEqual(result.status_code, 202)
+
+    def test_trip_same_area_under_10_km(self):
```
vinyll, line 47: 👍

vinyll (changes requested):

Ydrasil:
@vinyll

user/test.py
```diff
@@ -0,0 +1,55 @@
...
+        result = self.client.get('/privacy')
+        self.assertEqual(result.status_code, 200)
+
+    def test_delete_trip(self):
+        login_user(self)
+        lat_lng = [10, 10]
+        self.client.post('/trips/add', data=json.dumps(lat_lng),
+                         content_type='application/json')
+        user = User.objects.first()
+        url = '/privacy/[10, 10]/' + user.locations[0]\
```
vinyll: Ideally I would use the timestamp that is more url-friendly and unique enough:
```python
url = '/privacy/{}'.format(user.location[0].date.timestamp())
```

However, if for some reason you could not do that, at least formatting the url a little cleaner:

```python
url = '/privacy/10,10/{}'.format(user.location[0].date.strftime('%Y-%m-%d %H:%M:%S.%f'))
```

But really the first way may be more appropriate

user/test.py
```diff
@@ -0,0 +1,55 @@
...
+            .date.strftime('%Y-%m-%d %H:%M:%S.%f')
+        result = self.client.post(url, content_type='application/json')
```
vinyll: `.post()` that is actually deleting is confusing.
We probably want to do a `.delete()`.
If for some reason you couldn't (eg: you're using a form), you should have `/delete` in your url.

user/test.py
```diff
@@ -0,0 +1,55 @@
...
+        user = User.objects.first()
+        self.assertEqual(user.locations.count(), 0)
+        self.assertEqual(result.status_code, 200)
```
vinyll: Does it return something in the body? If not it should return a `204`

user/views.py
vinyll: `methods` should allow delete.
URL could be improved with just a timestamp (see previous comment).

vinyll (changes requested):
Sorry for long delay for review @Ydrasil.
Some url/method things to improve for deleting a ressource.

@fuentesloic might be interested into this feature and adding features to this frontend ;)

vinyll (approved):

PR #97, opened 2017-05-15
Fix bug that appear if the title or the content is empty

Ydrasil:


vinyll (commented):
This is a very good point.
However, how do these error desplay to the client? Are they displayed in the error banner or does it just break the server to the client (like an error 500)?

Ydrasil:
When the content is empty, its a server error `500`. Its because of the language detection that is trying to know what is the language is used in an empty content. After that the article appear correctly in the user's home page and we can correctly edit him. However the article doesn't seem to appear in `/articles`

When the title is empty, we can correctly save the article, he stay in the database, but we can't edit him. When we try to edit him (click on the article in the user's home page) a 404 error is appearing, and like the other bug above, the article doesn't seem to appear in `/articles`. This bug appear because we can't create a correct slug with an empty title.

vinyll:
I undestand that when there is no title or content, the server renders an error 500 when trying to display the article. Hence this PR makes sense.

However, my question is more on a user perspective. When the user is trying to save an article with no title or no content, he should see his article with an error message explaining that he must enter a title or a content to save his article.

I feel like `abort(400…)` would just display a crash page that is not helpful for the user. What if he had written an article and forgot the title? He should see his content again and say something like "Your article should have a title".

Ydrasil:
@vinyll What do you think about the error message now ?

blog/views.py
```diff
@@ -5,11 +5,12 @@
     Flask, request, render_template, redirect, url_for, abort, flash
 )
 from flask_login import current_user, login_required
+import langdetect
```
vinyll: why adding `langdetect` for the empty title or content?

blog/views.py
```diff
@@ -5,11 +5,12 @@
...
 
 from core import app, db
 from core.helpers import url_for_user, url_for_article
 from user.models import User
-from .models import Article
+from .models import Article, clean_html
```
vinyll, line 9: `clean_html` is a good point! 👍

blog/views.py
```diff
@@ -41,6 +42,9 @@ def article_create():
     if request.method == 'POST':
         article.title = request.form.get('title')
         article.content = request.form.get('content')
+        if article.title == '' or clean_html(article.content) == '':
+            return render_template('blog/article.html', article=article,
```
vinyll: A more conventional way to pass errors to a template is a variable called `errors` [like here](https://github.com/neomad-team/neomad.org/blob/master/auth/views.py#L28).

+1 for the _400_ status 👍

blog/views.py
```diff
@@ -55,11 +59,13 @@ def article_edit(id):
         abort(404)
     article.title = request.form.get('title')
     article.content = request.form.get('content')
+    if article.title == '' or clean_html(article.content) == '':
+        return render_template('blog/article.html', article=article,
+                               edit=True, empty=True), 400
```
vinyll: same here with the `errors`

templates/blog/article.html
```diff
@@ -29,6 +29,12 @@
     </div>
   </section>
 
+  {% if empty %}
```
vinyll: If you use `errors` as a list of errors, we could even extract [the specific errors rendering template portion](https://github.com/neomad-team/neomad.org/blob/master/templates/auth/login.html#L65) to move it into an [includable/reusable template portion](http://jinja.pocoo.org/docs/2.9/templates/#include)

vinyll (changes requested):

vinyll (approved):

PR #99, opened 2017-05-16
Add test for authentication

Ydrasil:


auth/test.py
```diff
@@ -0,0 +1,48 @@
...
+from user.views import *
+
+
+class AuthTest(TestCase):
+    def setUp(self):
+        self.client = app.test_client()
+        self.user = User(email='test@email.com').set_password('test').save()
+
+    def tearDown(self):
+        User.objects.delete()
```
vinyll: We should ensure we are running a test database with this command.
Not sure if we should go for documentation, default settings, or other solution

auth/views.py
```diff
@@ -25,7 +25,7 @@ def login():
             errors.append(invalid)
         else:
             errors.append(invalid)
-    return render_template('auth/login.html', errors=errors)
+    return render_template('auth/login.html', errors=errors), 401
```
vinyll: This is a very good point!
However you are returning an error 401 in every case here…
Maybe it would be interesting to cover this case with you tests, and then, once you can see that it fails (status code not being 200) when you just land on the page, you could adapt here ;)

vinyll (changes requested):
Small changes, but beautiful work on these tests!! 👏

auth/test.py
Ydrasil: I don't know either

auth/views.py
vinyll: Here's a proposal:
```python
return (render_template('auth/login.html', errors=errors),
        200 if not len(errors) else 401
```

vinyll: Here's a proposal:
```python
return (render_template('auth/login.html', errors=errors),
        200 if not len(errors) else 401
```

vinyll (approved):
Let's get things going on.

auth/test.py
vinyll: Maybe can use just deleted the created users?

auth/test.py
```diff
@@ -26,6 +26,40 @@ def test_signup(self):
         except User.DoesNotExist:
             self.fail('User does not exist')
         self.assertEqual(result.status_code, 302)
+        self.client.post('/signup', data=data)
+        # Sign up a new user with same slug for neomad
```
vinyll: Shouldn't this be a different test like `test_signup_duplicate_slug()`?

auth/test.py
```diff
@@ -26,6 +26,40 @@ def test_signup(self):
...
+            'email': 'newtest@test.com',
+            'password': 'test'
+        }
+        result = self.client.post('/signup', data=data)
+        try:
+            User.objects.get(username='newtest1')
+        except User.DoesNotExist:
+            self.fail('User does not exist')
+        self.assertEqual(result.status_code, 302)
+        # Sign up a new user with same slug for neomad twice in a row
```
vinyll: Same separated test question here

auth/test.py
```diff
@@ -26,6 +26,40 @@ def test_signup(self):
...
+            'email': 'newtest@testagain.com',
+            'password': 'test'
+        }
+        result = self.client.post('/signup', data=data)
+        try:
+            User.objects.get(username='newtest2')
+        except User.DoesNotExist:
+            self.fail('User does not exist')
+        self.assertEqual(result.status_code, 302)
+        # Sign up a new user that have a username that already exist
```
vinyll: again a new test

auth/test.py
Ydrasil: I have changed that

vinyll (approved):

PR #100, opened 2017-05-23
simplify around code

fuentesloic:


static/components/PoiCard.js
```diff
@@ -40,19 +40,19 @@ class PoiCard extends React.Component {
...
-        return `${distance} ${unit[0]}`
-      }
+    if(!distance){ 
+      return ''
+    } else if(distance > 1000) {
+      distance = Math.round(distance/1000)
+      return `${distance} ${unit[1]}`
+    } else {
+      return `${distance} ${unit[0]}`
     }
```
vinyll: What about:

```
renderDistance(distance, unit) {
  if(!distance) return ''
  distance = Math.round(distance/1000)
  return distance < 1000 ? `${distance} meters` : `${distance} km`
}
```

It seems more readable to me. What do you think?

fuentesloic: Oh I see ... so next level on writting javascript ^^
Sure ... that looks weird for now but I saw a lot of this "kind" of writting

vinyll: but the code is wrong.
Other suggestion:

```
renderDistance (distance) {
  if(!distance) return ''
  if(distance > 1000) {
    return `${Math.round(distance/1000)} km`
  }
  return `${distance} meters`
}
```

> I'm not using the `unit` array as I don't understand what it is needed for.

fuentesloic: I prefer this one !

vinyll (approved):

vinyll:
Please squash ;)

PR #101, opened 2017-05-23
added docker npm dependency and build task

vinyll:
The docker container should install npm (and [Yarn](https://yarnpkg.com/) for the sake of performance), and build the assets before running the project.

### DIFF
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -1,7 +1,14 @@
 #!/bin/bash
 
+# Add extra system dependencies
+apk --update add nodejs
+npm install yarn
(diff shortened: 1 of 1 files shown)


docker/entrypoint.sh
```diff
@@ -1,7 +1,14 @@
 #!/bin/bash
 
+# Add extra system dependencies
+apk --update add nodejs
```
vinyll, line 4: `nodejs` actually installs `npm`.

Ydrasil (approved):
Seems very clear to me :ok_hand:

PR #102, opened 2017-05-24
WIP : Around reactless

vinyll:
This is just a test to see how to compact the quantity and the spreading of the code.

Note that this is for testing purpose. I'm not sure we would go into this direction and it's very subject to reactions and opinions.

See [the appropriate commit](https://github.com/neomad-team/neomad.org/pull/102/commits/5bab044c93b352304c4e6e5267233d79ceea938d)

Note that we add 18 lines and remove 38. Meaning we cut half.


fuentesloic:
I agree with you it's very understandable ! just one question: why remove you `Power` & `Rank` react components and not `Comments` ??
Because `Power` & `Rank` don't have an evolution "build" by JS right ? (I want say a component that not use map(), Math() or any other function)

fuentesloic (commented):
I agree with you it's very understandable ! just one question: why remove you Power & Rank react components and not Comments ??
Because Power & Rank don't have an evolution "build" by JS right ? (I want say a component that not use `map()`, `fetch()`, `myFunction()` or any functions)

vinyll:
I just didn't delete the "comment" because it was 2:30am and less straight forward, and this was just a trial ;)

fuentesloic:
So ... if you delete `Comments` react component, you delete almost all the component, it's not a little weird ??

vinyll:
check out the title of this PR: "reactless" ;)

fuentesloic:
it's not false (that's work in english too ^^)

fuentesloic:
I prefer this version, reactless is better (even if I will writte in react again - maybe - )
So what do you think about squashing and merging it ?

vinyll:
Well, I'm not sure this is merge ready yet: it's still in React, but a bad React (because I broke the component you did like you should do).
Therefore I'd say either we go further on a road like this or we ditch this branch…

fuentesloic:
If we stay on the `react` way, we write again and again new feature and we'll finish to keep react because that will be a huge job ... but in other way `react` (looking like cool) do the job and is a new thing but I'm still thinking like you, react oblige people to have knowledge in react (very few 'cause it's easy to learn). 
Using preact is better because that oblige me to keep my react simple (I can't use rooter for exemple).
So it's just a choice, for me, in this projet, even if I like react, we have to swicht for vanillaJS

vinyll:
This may not be a good idea.
We'll reopen if ever.

PR #103, opened 2017-05-24
/trips - trip lines joining markers

fuentesloic:


vinyll:
hooo, I'm loving the idea of this feature ^^
Need to check out the code I guess ;)

static/js/trips-map.js
```diff
@@ -29,19 +34,50 @@ map.fitBounds([[
...
+                    'coordinates': lineCoords
+                }
+            }
+        },
+        'layout': {
+            'line-join': 'round',
+            'line-cap': 'round'
+        },
+        'paint': {
+            'line-color': 'rgba(0, 0, 255, 0.5)',
```
vinyll: Suggestion for a color and transparency: `rgba(212, 85, 85, 0.3)`.
I tried a few things and this seems to me like a good compromise.

vinyll (commented):
Nothing to change. I'm only arguing on the color and its transparency.
Could you check the color I tried and see if that suits you?
For the rest it's all great for me 👍

static/js/trips-map.js
fuentesloic: I think that too

PR #104, opened 2017-05-24
/around - responsive

fuentesloic:


vinyll (commented):
On a iPad the top padding of a card is broken:
screen shot 2017-05-25 at 8 25 12 am

On a mobile, it's quite unusable: there's no map but cards, and when clicking on a card, other cards disappear (I understand the technical reason, but it's unusable).

I would say to fallback displaying the map on mobile.
We could have the map on the upper half of the screen, and the cards on the lower half of the screen. When scrolling the cards, the map stays in place.
What do you think?

If that's too complex, I'd say we just display on a mobile like we display on a 600px… :/

fuentesloic:
that was not a problem of complexity more a probleme of my imagination 'cause a tried a lot of way without find out the good one. Thats sounds good for me so ... let's go :)

fuentesloic:
Large screen: pois card into two column
Middle screen: pois card into one column
Tablet screen: pois card into one column without all the information (comments is too huge)
Mobile screen: not pois card replacing by popup directly on the marker

static/js/around-map.js
```diff
@@ -70,7 +81,7 @@ map.on('click', event => {
 // functions
 function currentMarker (currentLatLng) {
   const popup = new mapboxgl.Popup({offset: [10, 0]})
-    .setText('Your current location')
+    .setHTML('<p>Your current location</p>')
```
vinyll: I would just say "<p>My current location</p>" instead, as we are using My/I/me on the project.

vinyll (commented):
Why 478px? Why not just 500px?

static/js/around-map.js
fuentesloic: sure ... I was pretty sure to change that a long time ago :)

fuentesloic:
why not :p ? (iphone 6+ size so a big screen for a mobile)

vinyll:
> why not :p ? (iphone 6+ size so a big screen for a mobile)

Because it seems more like a random value than an actual break-the-page layout.
I guess you should set a limit break by checking until what value the blocks still look nice, and rounding it up. @newick may have a better opinion about that.

fuentesloic:
> @newick may have a better opinion about that.

I agree with you ! @newick  heeeelp :)

vinyll (approved):
We really need to clarify the size of responsiveness.
I've been looking over the web and could eventually find this: http://www.websitedimensions.com

However, I once had read the concept of displaying each module (component-wise) until it has the appropriate space. Meaning that each module could have its responsive size, and I thought it made much more sense.

Nevertheless, I don't want to block this PR because of this, and I'd rather have @newick feedback than my non-css expertise.

fuentesloic:
I 100% agree !

newick:
@fuentesloic @vinyll 
Usually I use Bootstrap mediaqueries values 🙄
I tend to change that by a value defined by the design. If it starts to break at a random value, it could be the MQ value.

It seems to be the same idea as the last @vinyll 's one. 
I'll try to find some time this week to look after this problem

vinyll:
@newick: do you still want to have a look at this?

vinyll:
@fuentesloic merge?

newick:
sorry I didn't take time to rewrite my deleted comment!

It looks ok for me, juste instead of 70/30 for tablets i'll stay at 50/50.
There's a bug under something like 500-ish, the POI don't have height anymore

fuentesloic:
I want keep the 70/30 for tablets because I don't have enough space to write `comments` so I prefer have a little card with few important informations and a big map to navigate than cards with a bigger space available and a smallest card (that don't bring to users more informations)
But we could change this way/mind and discuss deepper about it after (like a PR with design/user XP subject).

fuentesloic:
@newick  I fixed the height ;)

PR #105, opened 2017-05-26
avatar menu

fuentesloic:


static/js/utils.js
```diff
@@ -17,7 +17,7 @@ function alert(type, message, delay) {
 
 window.onload = _ => {
   const menu = document.getElementById('menu')
-  const avatar = document.getElementById('avatarMenu')
+  const avatar = document.getElementById('avatar-img')
```
vinyll: Are you sure? https://github.com/neomad-team/neomad.org/blob/master/templates/layout.html#L23

fuentesloic: nop nop !!!!! I made a mistake

templates/layout.html
```diff
@@ -52,4 +53,4 @@
 
 {% include 'user/localize.html' %}
 
-{% include '_analytics.html' %}
+{% include '_analytics.html' %}
```
vinyll: You should keep the new line at the end of the file.
https://www.google.com/search?q=end+of+file+newline

And your editor should always have it automatically: https://stackoverflow.com/questions/43052279/atom-github-removing-end-of-file-newline

vinyll (changes requested):

templates/layout.html
fuentesloic: you fix it in pro right ?

vinyll:
ok, I re-added the newline in the end of the file.
Please @fuentesloic, make sure you set your editor to always have a newline in the end of your files (this is valid for all languages / all files).

vinyll (approved):

fuentesloic:
Sorry I just check (some friends at home ...) my settings was `whitespace: false` so I change it (And I don't know why that change, a new package I think !). Anyway, I'm back with `whitespace`

PR #106, opened 2017-06-01
Fix bug that appear when sign up with an already existing username

Ydrasil:


user/models.py
```diff
@@ -10,6 +11,24 @@
 from trips.models import UserLocation
 
 
+def change_username(username):
+    try:
+        # We stop when the exception is caught
+        while True:
+            str_number = re.match('.*?([0-9]*)$', username).group(1)
```
vinyll: Just a detail:
`str_number = re.match('.*([0-9]+)$', username).group(1)`

(to be tested though)

user/models.py
```diff
@@ -10,6 +11,24 @@
...
+            if str_number == '':
+                username = username + '1'
+            else:
+                number = int(str_number)
+                number = number + 1
```
vinyll: `number = int(str_number) + 1`

user/models.py
```diff
@@ -10,6 +11,24 @@
...
+                username = re.sub(r'\d+$', '', username)
```
vinyll: Why don't you extract it from your previous regex?

user/models.py
```diff
@@ -10,6 +11,24 @@
 from trips.models import UserLocation
 
 
+def change_username(username):
```
vinyll: I'd rather call this `uniquify_username` or similar if it changes the username.
or similar with `slug` if you go for the slug instead.

user/models.py
```diff
@@ -44,5 +63,12 @@ def __str__(self):
         return str(self.username or 'Unknown')
 
     def save(self, *args, **kwargs):
+        try:
+            User.objects.get(username=self.username)
```
vinyll, line 66: Notice that we don't really mind if the username is unique, but rather the slug (that's the one that crashes).

vinyll (changes requested):
That's a good point. I'd have thought of it but would have been too lazy to write it!
Good point 👍

user/models.py
Ydrasil: It will be trow an AttributeError if username does not contain any number

user/models.py
Ydrasil: I don't extract it from the previous regex because the previous regex only match the last digit

user/models.py
Ydrasil: The username is marked as unique and with the username we create the slug.
If I uniquify the slug, I have to uniquify the username again because it will not work. That's why I only uniquify the username.

vinyll: true, I didn't have unique username in mind.

user/models.py
```diff
@@ -10,6 +11,23 @@
 from trips.models import UserLocation
 
 
+def uniquify_username(username):
+    try:
+        # We stop when the exception is caught
+        while True:
+            str_number = re.match('.*?([0-9]*)$', username).group(1)
```
vinyll, line 18: Based on your idea, what do you think of the following?

```python
username, iterator = re.match('(.*)([0-9])?$', 'test').groups()
iterator = int(iterator) if iterator else 0
new_username = '{}{}'.format(username, iterator + 1)
User.objects.get(username=new_username)
```

user/models.py
```diff
@@ -44,5 +62,12 @@ def __str__(self):
         return str(self.username or 'Unknown')
 
     def save(self, *args, **kwargs):
+        try:
+            User.objects.get(username=self.username)
+            # If the code above is reach on the block try, it means that there
+            # is a duplicate
```
vinyll, line 68: BTW, this comment is also a good point as it's not straightfoward to understand. 👍

user/models.py
Ydrasil: With this solution, it will always add 1 to `username` because the regex match with `(.*)` and the digit at the end are included in `(.*)`. 
With a good regex it will be perfect (I'm still searching for a good one) :+1:

vinyll (approved):
I'm pretty certain that one regex should be able to extract the username and the iterator.
However, I'm bad at them and it might not be worth blocking this PR.

PR #108, opened 2017-06-09
Around - form to add poi

fuentesloic:


fuentesloic:
The form is not functionnal, you can "just" open or close it without bugs (I hope 🥇).
So, that's it, I want your mind before working more.

newick:
No time to pull it now :)
From the screenshots : 
- Why not use stars directly for the WiFi quality?
- Could be great to have a proper button on the left part to add POI, on the top or the bottom with a label

I'll have some times next week to speak about design or integration (or 🍺 )

static/css/around-map.css
```diff
@@ -68,6 +68,220 @@ main {
   align-self: flex-end;
 }
 
+/* poi form */
+
+.mapboxgl-canvas{
```
vinyll: space before curly bracket

static/css/around-map.css
```diff
@@ -68,6 +68,220 @@ main {
...
+#poi-form input[type=range]:focus::-ms-fill-lower {
+    background: #888;
+}
+
+#poi-form input[type=range]:focus::-ms-fill-upper {
+    background: #ccc;
+}
+
+#poi-form select option {
+  font-family: "Open Sans", Arial, Helvetica, sans-serif;
```
vinyll: why not a `font-family: inherit`?

static/css/around-map.css
```diff
@@ -68,6 +68,220 @@ main {
...
+}
+
+#close-poiForm {
```
vinyll: no camelcase

static/css/around-map.css
```diff
@@ -88,6 +302,10 @@ main {
   z-index: 1;
 }
 
+.display {
```
vinyll: a `display` class? You probably want a more explicit/consistent name

static/js/utils.js
```diff
@@ -16,6 +16,19 @@ function alert(type, message, delay) {
...
+  const displayForm = document.querySelectorAll('.display-form')
+  displayForm.forEach( button => {
+    button.addEventListener('click', _ => {
+      const poiForm = document.querySelector('#poi-form')
+      poiForm.classList.toggle('display')  
+    })
+  })
+  
   const menu = document.getElementById('menu')
   const avatar = document.getElementById('avatarMenu')
```
vinyll: no camelcase in html class name or html ids

templates/around/map.html
```diff
@@ -12,17 +12,68 @@
 {% block main %}
 
   <div id=wrap>
-    <div id=map></div>
+    <div id=map>
+      <button id=open-poiForm class=display-form type=button>
```
vinyll: no camel case name here

templates/around/map.html
```diff
@@ -12,17 +12,68 @@
...
+          </g>
+        </svg>
+      </button>
+    </div>
+    </div>
     <div id=poi-cards></div>
-  </div>
+    <div id=poi-form>
+      <form action= method=post>
+        <button id=close-poiForm class=display-form type=button>
```
vinyll: no camelcase id

templates/around/map.html
```diff
@@ -12,17 +12,68 @@
...
+          <li>
+            <label for=type>Type of place</label>
+            <select name=type>
+                <option value=coworking>Coworking space</option>
+                <option value=coffeeshop>Coffeeshop / Restaurant</option>
+                <option value=other>Library / Museum / other</option>
+            </select>
+          </li>
+          <li>
+            <label for=comment>Anything to say ?</label>
```
vinyll: no space before "?" in English

templates/around/map.html
```diff
@@ -12,17 +12,68 @@
...
+    </div>
 
   <script>
     var currentLatLng = {{ current_user.current_location or [] }}
     var locations = [
     {% for user in users %}
       {
         location: {{ user.current_location }},
-        user: "{{ user }}"
+        user: {{ user }}
```
vinyll: I don't think you tested this because JS is broken here.
`{{ user }}` is python, writing in a `<script>` section, therefore it's writing some JS.
this would generate `user: John` where _John_ would be undefined.
You want to keep the quotes `user: "John"`.
Avoid single quotes in this specific case as we may have single quotes in user names (much less chance to have double quotes though it could happen).

vinyll (changes requested):
I'm not confident about the browser specific pseudo selectors.
Also I share @newick's suggestions 👍

templates/around/map.html
fuentesloic: yes I don't try it ... I just remove the quote ... too many I think :(

fuentesloic:
@newick 
for the start: because I don't have a lot of time but it's a super cool idea
for the button: that was my first choice (on the top) but after I change because => responsive when we have only the card and also deal with a small place for card and a lot of information

vinyll:
What about one large button under the cards but always visible that would say "Add my spot" ?

templates/around/map.html
```diff
@@ -25,8 +25,8 @@
     </div>
     <div id=poi-cards></div>
     <div id=poi-form>
-      <form action="" method="post">
-        <button id=close-poiForm class="display-form" type="button">
+      <form action= method=post>
```
vinyll: `action= ` may not be valid. Keep the quotes in this case

fuentesloic:
but if we don't have the pois-card area (responsive) where your put the button ? on the map ?
A sign (like the one I use) is not understandable ?

vinyll:
I would then put the button on the map when the cards are not visible.
The current button was not obvious to me, but it might be only me. Other feedbacks are welcome :)

fuentesloic:
I tried like webnomad ( with a `+` sign but it's like zoom for me)

newick:
i understood it thanks to the hover effect. It doesn't exist on mobile.
The solution could be a button with a label on the bottom. For desktop it's down the left column, for mobile it's down hover the map ?

maybe ask @AymericFaivre for advice

fuentesloic:
New design proposition (working)
large: capture d ecran 2017-06-16 a 09 30 58
moyen: capture d ecran 2017-06-16 a 09 31 18
small: capture d ecran 2017-06-16 a 09 31 32

newick:
👍 
it just need, on desktop and tablet, some margin on the top, right and left of the button. Top must be the same as bottom and left/right must be equal with the POIs ;)

vinyll:
LGTM 🥇

static/js/around-map.js
```diff
@@ -176,13 +176,12 @@ function scrollCard (poi_id) {
 // close form
 const form = document.querySelector('#poi-form')
 form.addEventListener('click', event => {
-  if(form.classList[0] === 'visible') {
+  if(event.target === form) {
     form.setAttribute('class','')
```
vinyll: `event.target.classList.remove('visible')`

static/js/around-map.js
```diff
@@ -176,13 +176,12 @@ function scrollCard (poi_id) {
 // close form
 const form = document.querySelector('#poi-form')
 form.addEventListener('click', event => {
-  if(form.classList[0] === 'visible') {
+  if(event.target === form) {
```
vinyll: `event.target` is **always** equals to form.
You should just use `event.target`

vinyll (changes requested):

static/js/around-map.js
```diff
@@ -173,8 +173,22 @@ function scrollCard (poi_id) {
   window.scrollTo(0, (cardTop - cardHeight/2))  
 }
 
-// saving/adding a spot - form
+// form
+
+// display form with open/close button
+const form = document.querySelector('#poi-form')
+const displayForm = document.querySelectorAll('.display-form')
+displayForm.forEach(button => {
```
vinyll: displayForm = [button, button, button, …] ??

AymericFaivre:
Seems a good solution 👍 
In desktop, maybe you could put the things  in the button (or the background) in green to make it more obvious.

static/css/around-map.css
```diff
@@ -461,7 +464,7 @@ li > svg {
     padding: 0;
     background: transparent;
   }
-  
+
```
vinyll: remove the extra space

static/js/around-map.js
```diff
@@ -173,8 +173,22 @@ function scrollCard (poi_id) {
...
+  button.addEventListener('click', _ => {
+    form.classList.toggle('visible')
+  })
+})
+// close form clicking outside form
+form.addEventListener('click', event => {
+  event.target.classList.remove('visible')
```
vinyll: `click` on a `form`?

vinyll (changes requested):

static/js/around-map.js
fuentesloic: yes I have two buttons to display with the same fonction so with `forEach` I give to both the `AddEventListener`

vinyll:
#145 introduces a new version of the API.
We should adapt the form to this version.

fuentesloic:
@vinyll why you re put it on WIP ?? for me it's working (but not test possible without my local db)

fuentesloic:
@vinyll: For my this PR done and working

vinyll:
I've tested and we're definitely very close.
However the wifi stars seems not to work.

For the following data:
```
name | "test me"
location |  
0 | 48.8569297790527
1 | 2.44120011329651
wifi | 2
power | true
category | "other"
comments |  
0 | "Just a test"
creation_date | "1503242050.133"
id | "5999c3629bc97c22d047317d"
user_url | "/api/users/5914dbdca193770016e289a7/"
```

I have this result:

screen shot 2017-08-20 at 7 19 20 pm

static/css/around-map.css
```diff
@@ -116,6 +116,8 @@ footer {
   background: rgba(0, 0, 0, 0.6);
   z-index: 2;
   overflow: scroll;
+  z-index: 10;
```
vinyll: `z-index` is already mentioned 2 lines above.
And why 10`?

vinyll (approved):
Good for when `z-index` is fixed

static/css/around-map.css
fuentesloic: @vinyll 10 to be stronger than the header witch need a 9 to be stronger than the map (thank you OSM ^^)

vinyll: When something so specific comes up, just give the code a comment explaining why.  Like `/* Above header */`

PR #109, opened 2017-06-12
The save bar isn't hidden by the footer anymore

Ydrasil:


Ydrasil:
@vinyll

newick (commented):
instead of removing the `position: fixed` maybe it could be a better idea to fix it just over the footer ?
With `bottom: 87px`it works :)

Ydrasil:
I don't know if its a good idea, because if we change the footer, then we'll must refixed the save bar again.

newick:
yeah, but the design looks better like this ;)

vinyll:
Does the bar still appear in the bottom of the screen even if the article is longer than the height of the screen?

screen shot 2017-06-15 at 11 16 00 am

I'm surprised we had a `position: fixed` that was just useless.

vinyll:
My guess would be that we want to position to be at the bottom of the screen with long article (so that the bar is constantly visible), and above the footer when it's visible.
I think the css position `sticky` is supposed to do the magic.

Ydrasil:
Something like this @vinyll ?

newick:
Do we count Edge users as our users? 
http://caniuse.com/#feat=css-sticky

vinyll:
@Ydrasil I guess something like that, but it should be manually tested to see if the behavior is the one expected (which is not the case with the current `position: sticky` code).

screen shot 2017-06-15 at 11 37 47 am

Ydrasil:
If the save bar have to be always overhead the footer then it works

vinyll:
> Do we count Edge users as our users?
> http://caniuse.com/#feat=css-sticky

I would say _let's make it easy_.
This is an editing feature, I suppose we can limit editing to bleeding edge browsers or FF/Chromium. (Edge 16 will support).
I would be more hesitant for reading the website.

newick:
ok for me then ;)

fuentesloic:
I put a Z-index stronger than the one on footer, so like that, the save-bar still stay on the footer
(ps: sorry for the dirty commit name, I write that directly on github, so I could squash all of them when that'll be validate)

vinyll (approved):
I believe the `z-index` of the bar is the default (0). Therefore you could use `1` as value.
If you confirm it works with `1`, just change it and please squash the 2 commits (as you undid the 1st one).

fuentesloic:
a `z-index: 1` is on the footer, so I have to put `2`

Issue #27, opened 2017-03-31
Article edit meta data


vinyll:
An article has meta data such has a status (draft/published), a slug, a publication date, …
We should find a way to present the editable meta data form within the article page.

vinyll:
Step 1: _publication status_.

Try to keep in mind that we may need other meta data editing.

PR #115, opened 2017-06-26
Add publication status for articles

Ydrasil:
Now we have to think about how to implement this feature for user experience
Cc @vinyll


blog/test.py
```diff
@@ -38,3 +49,18 @@ def test_language_detection(self):
...
+
+    def test_article_appear_by_default_in_articles(self):
+        article = Article(title='<h1>title<br></h1>', content='<p>content</p>')
+        article.author = self.user
+        article.save()
+        result = self.client.get('/articles')
+        self.assertIn(b'<article class=preview>', result.data)
+
+    def test_unpublished_article_does_not_appear_in_articles(self):
+        article = Article(title='<h1>title<br></h1>', content='<p>content</p>')
```
vinyll: That's a bug you are raising here. We should not have `<br>` in the title

blog/models.py
```diff
@@ -39,6 +39,7 @@ class Article(db.Document):
     author = db.ReferenceField(User)
     language = db.StringField(min_length=2, max_length=2, default='en')
     images = db.ListField()
+    is_published = db.BooleanField()
```
vinyll: I guess we could use a `publication_date` instead of a `is_published`.

`is_published` could then be a property:

```python
@property
def is_published():
    return bool(self.publication_date)
```

user/test.py
```diff
@@ -0,0 +1,46 @@
...
+    def setUp(self):
+        self.client = app.test_client()
+        self.user = User(email='emailtest@test.com').set_password('testtest')
+        self.user.save()
+
+    def tearDown(self):
+        User.objects.delete()
+        Article.objects.delete()
+
+    def test_unpublished_articles_does_not_appears_if_the_author_not_logged(
```
vinyll: I believe you could move any test about article in the _blog_ module.
Let's imagine the project would someday see the _blog_ module removed,  these tests should be removed as well. (not planned hu, just exposing modules responsibilities).

user/test.py
```diff
@@ -0,0 +1,46 @@
...
+            self):
+        article = Article(title='<h1>Article that must not appear<br></h1>',
+                          content='<p>content</p>')
+        article.author = self.user
+        article.is_published = False
+        article.save()
+        result = self.client.get('/@none')
```
vinyll: Should we actually have a user with _@none_ as slug?

vinyll (changes requested):
That's a well written feature with tests, good job again 👍 
I believe it would be better if you were working with @fuentesloic for implementing the frontend as well.

It wouldn't make sense to implement _dead code_ (code that is not used); we try to remove any code that is not used! ;)

blog/models.py
Ydrasil: I have a question about this.
I'll explain this with an example.
If I publish an article, then I unpublish this article, then I publish the article again, what will be the date of publication ? The first time I've published it or the last time ?

vinyll:
@fuentesloic: Shall we make an appointment to work together with @Ydrasil?

user/test.py
```diff
@@ -1,19 +1,33 @@
...
+    }
+    self.client.post('/login', data=data, follow_redirects=True)
+    
+
 class UserTest(TestCase):
     def setUp(self):
         self.client = app.test_client()
         self.user = User(email='emailtest@test.com').set_password('testtest')
+        self.user.allow_localization = True
         self.user.save()
```
vinyll:
```python
self.user = (User(email='emailtest@test.com',
                allow_location=True)
                .set_password('testtest')
                ).save()
```

user/test.py
```diff
@@ -44,3 +58,28 @@ def test_unpublished_articles_does_appears_if_the_author_is_logged(self):
...
+
+    def test_user_page_that_does_not_exist(self):
+        result = self.client.get('/@doesnotexist')
+        self.assertEqual(result.status_code, 404)
+
+    def test_privacy_page_unauthorized(self):
+        result = self.client.get('/privacy')
+        self.assertEqual(result.status_code, 401)
+
+    def test_privacy_page(self):
```
vinyll: a more explicit name like `def test_privacy_page_access(self)`?

vinyll (commented):
Will be good for merge after the small code improvements and the frontend implementation :)

blog/models.py
vinyll: @Ydrasil I would say it's the first day of publication.
We have the `modification_date` for other purposes.

fuentesloic:
I'm on it!
Sorry for the late! :)

templates/blog/macros.html
```diff
@@ -17,10 +17,12 @@
       {% else %}
       <time datetime={{ article.creation_date }} pubdate>{{ article.creation_date.strftime('%d %B, %Y') }}</time>
       {% endif %}
+      {% if edit %}
       <form>
         <input name="published" id="published" type="checkbox">
```
vinyll: no quotes required here

templates/blog/macros.html
```diff
@@ -17,10 +17,12 @@
...
         <label for="published" title="draft or publish">published</label>
```
vinyll: neither here

vinyll (changes requested):

templates/blog/macros.html
```diff
@@ -19,8 +19,8 @@
       {% endif %}
       {% if edit %}
       <form>
-        <input name="published" id="published" type="checkbox">
-        <label for="published" title="draft or publish">published</label>
+        <input name=published type=checkbox>
```
vinyll: you removed the `id` but the label's `for` was refering to it

templates/user/profile.html
```diff
@@ -55,8 +55,8 @@ <h1 name=username>{{ user.username or 'Insert a username' }}</h1>
     <p><span class=newtext>&nbsp;</span><span class=newtext>&nbsp;</span><span class=newtext>&nbsp;</span></p>
     <a href={{ url_for('article_create') }} class=read>Write this article</a>
     <form>
-      <input name="published" id="published" type="checkbox">
-      <label for="published" title="draft or publish">published</label>
+      <input name=published type=checkbox>
+      <label for='published' title=draft or publish>published</label>
```
vinyll: no need of quotes, even single quotes.
The `for` links to an `id`. You just deleted the id…

vinyll (changes requested):

templates/user/profile.html
```diff
@@ -73,6 +73,12 @@ <h1 name=username>{{ user.username or 'Insert a username' }}</h1>
     )
     apiFeedback(call, 'Your username has been updated.')
   })
+  
+  const publishField = document.querySelector('[name=published]')
+  publishField.addEventListener('change', event => {
+    apiFeedback(call, 'Your article is publish preference has been updated.')
```
vinyll: It could be nice to have a message that mentions the state:
"You article has been published." or "You article has been unpublished."

templates/user/profile.html
```diff
@@ -73,6 +73,12 @@ <h1 name=username>{{ user.username or 'Insert a username' }}</h1>
...
+  })
+  publishField.addEventListener('change', event => event.target.checked)
```
vinyll: What is this supposed to do?

templates/user/profile.html
```diff
@@ -54,10 +54,6 @@ <h1 name=username>{{ user.username or 'Insert a username' }}</h1>
     <h1><span class="newtext title">&nbsp;</span></h1>
     <p><span class=newtext>&nbsp;</span><span class=newtext>&nbsp;</span><span class=newtext>&nbsp;</span></p>
```
vinyll: WTF is this?

templates/user/profile.html
vinyll: What is this supposed to do?

templates/user/profile.html
vinyll: We could display a message that would be contextual:

'You article has been published.'
'You article has been unpublished.'

vinyll (changes requested):

user/views.py
```diff
@@ -14,10 +14,14 @@
 def profile(username):
     try:
         user = User.objects.get(slug=username)
+        if user == current_user:
+            articles = Article.objects(author=user)
+        else:
+            articles = Article.objects(author=user, is_published__ne=False)
```
vinyll:
```python
articles = Article.objects(author=user)
if user == current_user:
    articles = articles.filter(is_published__ne=False)
```

Should do the trick in a shorter way ;)

user/test.py
```diff
@@ -52,3 +56,21 @@ def test_delete_trip(self):
...
+        Article(title='<h1>Article that must not appear<br></h1>',
+                content='<p>content</p>', author=self.user,
+                is_published=False).save()
+        result = self.client.get('/@emailtest')
+        self.assertNotIn(b'Article That Must Not Appear', result.data)
+        self.assertEqual(result.status_code, 200)
+
+    def test_unpublished_articles_does_appears_if_the_author_is_logged(self):
+        login_user(self)
+        Article(title='<h1>Article that must appear<br></h1>',
```
vinyll: no html please, it must be cleaned out anyways (and you have a dedicated test for that) ;)

user/test.py
```diff
@@ -52,3 +56,21 @@ def test_delete_trip(self):
...
+                content='<p>content</p>', author=self.user,
+                is_published=False).save()
+        result = self.client.get('/@emailtest')
```
vinyll: ending slash

vinyll (changes requested):
You may also want to test to access an article with it's url if it's **un**published.

vinyll:
We will still probably need the publication date.
It may be a good option to replace `is_published` with `publication_date` and have it to `None` when it's unpublished.

templates/blog/macros.html
```diff
@@ -33,6 +33,7 @@
         <img src="{{ article.image | default('/static/images/articles/default.png', true) }}" width=100%>
       </figure>
       {{ article.content|striptags|truncate(200) }}
+    {% set url = url_for_article(article) %}
     <a href={{ url }} lang={{ article.language }} class='read action-button'>Read more</a>
```
vinyll: Using `{% set url = url_for_article(article) %}` is great when you need to reuse `url` multiple times.

Here you could just say: `<a href={{ url_for_article(article) }}`

templates/user/profile.html
```diff
@@ -82,7 +82,7 @@ <h1 name=username>{{ user.username or 'Insert a username' }}</h1>
 
   const publishField = document.querySelector('[name=publish]')
   publishField.addEventListener('change', event => {
-    const call = api.send('post', "{{ url_for('profile_edit') }}",
+    const call = api.send('post', event.target.form.action,
```
vinyll: 👍

templates/blog/article.html
```diff
@@ -60,11 +60,15 @@ <h1 name=title>{{ article.title or 'Insert your title here' }}</h1>
...
 <script src=https://cdn.rawgit.com/orthes/medium-editor-insert-plugin/3.0/dist/js/medium-editor-insert.min.js></script>
 
-{% set url = url_for('article_edit', id=article.id) if article.id else url_for('article_create')  %}
+{% set url = url_for('article_edit', id=article.id) if article.id else url_for('article_create') %}
 <form method=post action={{ url }} edit>
   <div class=toolbar>
     <input type=hidden name=title>
     <input type=hidden name=content>
+    <div>
+      <input name=publish type=checkbox>
```
fuentesloic: need an id

templates/blog/macros.html
```diff
@@ -17,13 +17,23 @@
       {% else %}
       <time datetime={{ article.creation_date }} pubdate>{{ article.creation_date.strftime('%d %B, %Y') }}</time>
       {% endif %}
+      {% if edit %}
+      {% set url = url_for('article_edit', id=article.id) %}
+      <form method=post action={{ url }} edit>
+        <div>
+          <input name="publish" id="publish" type="checkbox">
```
fuentesloic: need a specific id (with the article.id maybe)

templates/user/profile.html
```diff
@@ -79,6 +80,15 @@ <h1 name=username>{{ user.username or 'Insert a username' }}</h1>
   })
   localizeField.addEventListener('change', event => event.target.checked && localize())
 
+  const publishField = document.querySelector('[name=publish]')
```
fuentesloic, line 83: use querySelector with id

templates/blog/macros.html
fuentesloic: oh true ! my bad

templates/user/profile.html
vinyll: why a querySelecot with id?

templates/user/profile.html
fuentesloic: ;) I learned with someone pretty good 💯 / 💯

templates/user/profile.html
fuentesloic: because if I use by `name` that just take the firstbecause if I use by `name` that just take the first

vinyll: Do you have multiple time the `name=publish` in your document?

fuentesloic: yes on each article in you profile

fuentesloic: but don't succeed to fix it ... in each case the ID block me because I want a string without number or sign

vinyll: You're not publishing from your profile but from the article page, right?

vinyll: mhhh… I see we're publishing an article from templates/user/profile.html.
Why not publishing an article from the article's page?

fuentesloic: when you create an article it's right
But after that, I was wondering that user could be publish/unpublish article from is profile view

vinyll: Well the issue you're on is supposed to solve [editing article's meta data](https://github.com/neomad-team/neomad.org/issues/27).
Changing the status of an article would be interested to be done in a list if a user needed to have a massive status change for multiple articles, but the purpose here is to be able to change non visible informations of an article (like the publication status, publication date, and later other infos such as the slug, transfer the owner and so on).
Hence the interest of having a status bar.

I'm running this version on my local copy to check the points of views ;)

vinyll:
As the owner of the article https://dev.neomad.org/@vinyll/developpeur-sans-domicile-fixe-58faa70b1675bb01b46fcfa4/, when editing my article, I have a toolbar in the bottom for saving or deleting my article.
It could be a good place where to offer extra options, like (un)publishing, or changing the publication date.

screen shot 2017-08-23 at 10 22 12 pm

We could have either the easy version (replacing the _save_ button with 2 buttons: _save as draft_ and _save and publish_), or an icon featuring a more advanced sub-menu with options (like we have for the user menu, see the screenshot below)


screen shot 2017-08-23 at 10 27 14 pm
)

fuentesloic:
@vinyll it's already the case

templates/user/profile.html
vinyll: I've completed user tests with the following feedback: https://github.com/neomad-team/neomad.org/pull/115#issuecomment-324452994

templates/blog/article.html
```diff
@@ -66,7 +66,8 @@ <h1 name=title>{{ article.title or 'Insert your title here' }}</h1>
     <input type=hidden name=title>
     <input type=hidden name=content>
     <div>
-      <input name=publish type=checkbox>
+      {% set id=article.id %}
+      <input name=publish type=checkbox class=publish id=publish{{ id }}>
```
vinyll: is this id thing still useful?

templates/blog/macros.html
```diff
@@ -19,7 +19,8 @@
       {% set url = url_for('article_edit', id=article.id) %}
       <form method=post action={{ url }} edit>
         <div>
-          <input name="publish" id="publish" type="checkbox">
+          {% set id=article.id %}
+          <input name=publish type=checkbox class=publish id="publish{{ id }}">
```
vinyll: same question

blog/test.py
```diff
@@ -165,3 +165,18 @@ def test_language_detection(self):
...
+        article.save()
+        result = self.client.get('/articles')
+        self.assertIn(b'<article class=preview>', result.data)
+
+    def test_unpublished_article_does_not_appear_in_articles(self):
+        article = Article(title='title', content='<p>content</p>')
+        article.is_published = False
+        article.author = self.user
+        article.save()
+        result = self.client.get('/articles')
```
vinyll: ending slash

blog/test.py
```diff
@@ -165,3 +165,18 @@ def test_language_detection(self):
                           content='<p>The content of the article is in '
                                   'English</p>').save()
         self.assertTrue(article.language, 'en')
+
+    def test_article_appear_by_default_in_articles(self):
+        article = Article(title='title', content='<p>content</p>')
+        article.author = self.user
+        article.save()
+        result = self.client.get('/articles')
```
vinyll: ending slash

templates/blog/article.html
```diff
@@ -60,12 +60,14 @@ <h1 name=title>{{ article.title or 'Insert your title here' }}</h1>
...
 
-{% set url = url_for('article_edit', id=article.id) if article.id else url_for('article_create')  %}
+{% set url = url_for('article_edit', id=article.id) if article.id else url_for('article_create') %}
 <form method=post action={{ url }} edit>
   <div class=toolbar>
     <input type=hidden name=title>
     <input type=hidden name=content>
-    <button type=submit class=save>Save</button>
+    <input type=hidden name=is_published>
+    <button type=submit class=save value="">Save as draft</button>
```
vinyll, line 75: change the color of draft for a gray background?

templates/blog/macros.html
```diff
@@ -22,7 +22,7 @@
         <img src="{{ article.image | default('/static/uploads/articles/default.png', true) }}" width=100%>
       </figure>
       {{ article.content|striptags|truncate(200) }}
-    <a href={{ url }} lang={{ article.language }} class='read action-button'>Read more</a>
+    <a href={{ url_for_article(article) }} lang={{ article.language }} class='read action-button'>Read more</a>
```
vinyll: double quotes for html when quotes are required

user/test.py
```diff
@@ -52,4 +56,23 @@ def test_delete_trip(self):
         result = self.client.post(url, content_type='application/json')
         user = User.objects.first()  # DB was updated.
         self.assertEqual(user.locations.count(), 0)
-        self.assertEqual(result.status_code, 302)
+        self.assertEqual(result.status_code, 204)
+
+    def test_unpublished_articles_does_not_appears_if_the_author_not_logged(
+            self):
+        Article(title='<h1>Article that must not appear<br></h1>',
```
vinyll: remove html in title

user/test.py
```diff
@@ -52,4 +56,23 @@ def test_delete_trip(self):
...
+                content='<p>content</p>', author=self.user,
+                is_published=False).save()
+        result = self.client.get('/@emailtest')
```
vinyll: ending slash

vinyll (changes requested):

vinyll:
This PR seems almost good.
The only thing would be to store this `publication_date` instead of the `is_published` as a boolean.

When the article is published, we should save the current (`datetime.datetime.utcnow()`) date as publication_date.
When the article is saved as draft, the publication_date should be `None`.

templates/blog/macros.html
fuentesloic: so href requires quotes I think it's the same for `src` in `img` (isn't?)

vinyll: yes, the rule is simple:

1. don't use quotes unless required
2. when required, use double quotes

fuentesloic: how you know if it's require ? like I know for class I need it when we have a space inside ... same for src.
There are other attribute ?

vinyll: You know the rule, same for any attribute: if there's a caracter that would break the html in the value of the attribute, you need quotes. Such as a space, a `=`, a `<`, ` >, ` `/`, a `'` …

fuentesloic:
I push some design stuff here because that was quick (and I add condition into home previews article):
- When you edit:
capture d ecran 2017-08-25 a 10 31 16-fullpage
- In your article list:
capture d ecran 2017-08-25 a 10 47 24-fullpage

vinyll:
The draft ribbon seems appropriate to me. Just wondering about design details: should we go for flat design and remove the linear gradient? Should we make the ribbon width thinner?

fuentesloic:
@vinyll I should change design for be closer to the actual design (with color). Let me do an other try.

vinyll:
Right. That will probably get into the styleguide that @newick is working. Which – by the way – is going to shake some ass!!! 🚀

fuentesloic:
New flatty design purpose, font lighter, using neomad link color
capture d ecran 2017-08-25 a 13 50 41-fullpage

fuentesloic (dismissed):

blog/test.py
```diff
@@ -165,3 +168,17 @@ def test_language_detection(self):
...
+                author=self.user).save()
+        result = self.client.get('/articles/')
+        self.assertIn(b'A title for an article', result.data)
+
+    def test_unpublished_article_does_not_appear_in_articles(self):
+        Article(title='title', content='<p>content</p>',
+                publication_date=None,
+                author=self.user).save()
+        result = self.client.get('/articles/')
+        self.assertNotIn(b'<article class=preview>', result.data)
```
vinyll, line 184: I'd rather test that the title of the article does not appear, just like you do in line 177.

blog/views.py
```diff
@@ -11,7 +13,7 @@
 
 @app.route('/articles/')
 def article_list():
-    articles = Article.objects.all()
+    articles = Article.objects.filter(publication_date__ne=None)
```
vinyll: You probably want to go for something like http://docs.mongoengine.org/guide/querying.html#default-document-queries.

blog/views.py
```diff
@@ -46,6 +48,10 @@ def article_create():
         else:
             article.title = request.form.get('title')
             article.content = clean_html(request.form.get('content'))
+            if request.form.get('publish') == 'on':
```
vinyll, line 51: Relaying on the value `'on'` is pretty arbitrary as the value may change from the frontend when the frontend would evolve.
You should just: `if bool(request.form.get('publish')):…`
The responsability would be for the frontend to be coherent and send something that is [truthy of falsy](https://developer.mozilla.org/en/docs/Glossary/Truthy)

core/views.py
```diff
@@ -10,5 +10,5 @@
 
 @app.route('/')
 def home():
-    articles = Article.objects.all()[:3]
+    articles = Article.objects.all()[:3].filter(is_published__ne=False)
```
vinyll: http://docs.mongoengine.org/guide/querying.html#default-document-queries would factorize your code as above

static/css/_article_list.css
```diff
@@ -56,3 +56,48 @@
...
+  color: #fff;
+  text-align: center;
+  transform: rotate(45deg);
+  position: relative;
+  padding: .7rem 0;
+  left: -0.5rem;
+  top: 1.5rem;
+  width: 12rem;
+  background: #279E6A;
+  box-shadow: 0 0 0.3rem rgba(0,0,0,0.3);
```
vinyll: spaces after commas

static/css/edit.css
```diff
@@ -152,3 +152,32 @@ main > header figure .loading.ready::before {
...
+  color: #fff;
+  text-align: center;
+  transform: rotate(45deg);
+  position: relative;
+  padding: .7rem 0;
+  left: 1.5rem;
+  top: 1.5rem;
+  width: 16rem;
+  background: #279E6A;
+  box-shadow: 0 0 0.3rem rgba(0,0,0,0.3);
```
vinyll: spaces after commas

static/css/edit.css
```diff
@@ -152,3 +152,32 @@ main > header figure .loading.ready::before {
...
+  text-transform: capitalize;
+  color: #fff;
+  text-align: center;
+  transform: rotate(45deg);
+  position: relative;
+  padding: .7rem 0;
+  left: 1.5rem;
+  top: 1.5rem;
+  width: 16rem;
+  background: #279E6A;
```
vinyll: lower case or uppercase? Choice is yours, but stick with it ;)

templates/blog/article.html
```diff
@@ -39,6 +39,12 @@
     </ul>
   {% endif %}
 
+  {% if article.is_published==false %}
```
vinyll: `{% if not article.is_published %}` is more read-friendly ;)

templates/blog/macros.html
```diff
@@ -15,14 +15,19 @@
       {% else %}
       <time datetime={{ article.creation_date }} pubdate>{{ article.creation_date.strftime('%d %B, %Y') }}</time>
       {% endif %}
+      {% if not article.publication_date %}
```
vinyll: {% if not article.is_published %}

user/test.py
```diff
@@ -48,8 +54,25 @@ def test_delete_trip(self):
...
+        url = '/privacy/{}/delete/'.format(user.locations[0].date.timestamp())
         result = self.client.post(url, content_type='application/json')
+        self.assertEqual(result.status_code, 204)
         user = User.objects.first()  # DB was updated.
         self.assertEqual(user.locations.count(), 0)
-        self.assertEqual(result.status_code, 302)
+
+    def test_unpublished_articles_does_not_appears_if_the_author_not_logged(
+            self):
+        Article(title='<h1>Article that must not appear<br></h1>',
```
vinyll: title with html??

user/views.py
```diff
@@ -14,10 +14,14 @@
 def profile(username):
     try:
         user = User.objects.get(slug=username)
+        if user == current_user:
+            articles = Article.objects(author=user)
+        else:
+            articles = Article.objects(author=user, publication_date__ne=None)
```
vinyll:
```python
articles = Article.objects(author=user)
if user != current_user:
    articles.filter(publication_date__ne=None)
```
Though refactoring with a queryset method would change this

vinyll (changes requested):

templates/blog/article.html
fuentesloic: ^^

vinyll:
Also bewarn of the commits. A manual interactive rebase might be appropriate here:
- for preserving who did what
- for renaming commits (some are still with a _WIP_ – work in progress – which obviously don't make sense for an old commit)
- for reducing 40something commits into a few
- for removing titles that don't make sense without their context

Ydrasil:
I have two idea for rebase interactively this branch, squash all the commit into one feature that is the publication status, or by keeping 1 commit with backend and 1 other for frontend.
What do you think about this solution and have you another one ?

vinyll:
If I were you, I would go for 1 or a couple of commit for each – the backend and the frontend – as you say, to preserve what each developer has done.

fuentesloic:
@Ydrasil could you squash commit plz ? (and I think you need a rebase) I would like merge it on master

vinyll (approved):

PR #122, opened 2017-06-28
clean utils.js

fuentesloic:


static/js/around-map.js
```diff
@@ -173,7 +173,14 @@ function scrollCard (poi_id) {
   window.scrollTo(0, (cardTop - cardHeight/2))  
 }
 
-// saving/adding a spot - form
+// pois form
+const displayForm = document.querySelectorAll('.display-form')
+displayForm.forEach( button => {
```
vinyll: Still a mistery why `.display-form` is not a form but a list of clickable buttons.
Could you rename so it's more explicit?

vinyll (changes requested):
Did you check on different pages that the menu still works everywhere?

static/js/around-map.js
fuentesloic: because the button to open or to close have a class name `.display-form`
So I use it to toggle a class `visible`on the form
But I'll change

fuentesloic:
I test and that works

templates/around/map.html
```diff
@@ -14,7 +14,7 @@
   <div id=wrap>
     <div id=map></div>
     <div id=cards-wrap>
-      <button id=open-form class=display-form type=button>
+      <button id=open-form type=button>
```
vinyll: Why `type=button`?

templates/around/map.html
```diff
@@ -27,7 +27,7 @@
     </div>
     <div id=poi-form>
       <form action="{{ url_for('spot_create') }}" method=post>
-        <button id=close-form class=display-form type=button>
+        <button id=close-form type=button>
```
vinyll: Same question here

templates/around/map.html
fuentesloic: for nothing it's a mistake and it's useless

templates/around/map.html
fuentesloic: same mistake here

PR #124, opened 2017-07-01
read more from the same author at the end of an article

vinyll:
This is just a first step.
In a future PR I'd like to implement latest article preview from the same author.

screen shot 2017-07-01 at 6 30 57 pm

### DIFF
--- a/static/css/article.css
+++ b/static/css/article.css
@@ -44,3 +44,15 @@ article .date {
   margin-top: 3em;
   font-size: .7em;
 }
+
+article aside a {
+  display: block;
+  margin: 1em auto;
+  width: 400px;
+  padding: 1em 2em;
+  background-color: #79d;
+  color: white;
+  text-align: center;
+  font-weight: bold;
+  margin: 1em;
+}
--- a/static/css/article.css
+++ b/static/css/article.css
@@ -2,8 +2,13 @@ html {
   background-color: #fff;
 }
 
+main {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+}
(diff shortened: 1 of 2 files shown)


static/css/article.css
```diff
@@ -44,3 +44,15 @@ article .date {
   margin-top: 3em;
   font-size: .7em;
 }
+
+article aside a {
```
fuentesloic: according your HTML it's just `aside a` not `article aside a`

templates/blog/article.html
```diff
@@ -46,8 +46,7 @@ <h1 name=title>{{ article.title or 'Insert your title here' }}</h1>
 </article>
 
 <aside>
```
fuentesloic: You should remove `aside` when author are editing (and reading?) is own articles

fuentesloic (changes requested):

fuentesloic:
also you could applied on `main` a flex comportement to center the article and the aside
```
main {
  display: flex;
  flex-direction: column;
  align-items: center;
}
```
then remove the useless `margin` in `article`

static/css/article.css
```diff
@@ -44,3 +44,15 @@ article .date {
...
+  display: block;
+  margin: 1em auto;
+  width: 400px;
```
fuentesloic: please use `rem` to be consistant :p

fuentesloic (changes requested):
You should add a specific `:hover` 'cause it's green on the `color`

fuentesloic (approved):

PR #131, opened 2017-07-18
Add basic search feature

Ydrasil:


search/test.py
```diff
@@ -0,0 +1,56 @@
...
+from search.views import *
+from user.models import User
+from core import app
+
+
+class SearchTest(TestCase):
+    def setUp(self):
+        self.client = app.test_client()
+
+    def tearDown(self):
```
vinyll, line 13: This copy paste may mean it's time to merge some code.
You may want to create something similar to `core.Testcase` and extend it.

search/test.py
```diff
@@ -0,0 +1,56 @@
...
+        User.objects.delete()
+        Article.objects.delete()
+
+    def test_search_retrieve_several_user(self):
+        User(email='john@test.com',
+             username='john').set_password('search').save()
+        User(email='johndoe@test.com',
+             username='johndoe').set_password('search').save()
+        response = self.client.get('/search/john')
```
vinyll: We probably want to call the _GET_ with a search in a query parameters: `/search/?q=john`

search/test.py
```diff
@@ -0,0 +1,56 @@
...
+        # Verify that there is 2 distinct users in response
+        self.assertEqual(len(result['users']), 2)
+        self.assertEqual(len(result['articles']), 0)
+        self.assertIn('john', json.loads(result['users'][0])['slug'])
+        self.assertIn('john', json.loads(result['users'][1])['slug'])
+        self.assertNotEqual(result['users'][0], result['users'][1])
+
+    def test_search_retrieve_several_article(self):
+        Article(title='<h1>Article<br></h1>',
+                content='Why my french fries are cold ?').save()
```
vinyll: remove space before _?_ ;)

search/views.py
```diff
@@ -0,0 +1,17 @@
...
+
+@app.route('/search/<string:term>')
+def search(term):
+    regex = re.compile('.*{}.*'.format(term))
+    users = User.objects(slug=regex)
+    articles = Article.objects(content=regex)
+    result = dict()
+    result['users'] = [user.to_json() for user in users]
+    result['articles'] = [article.to_json() for article in articles]
+    return json.dumps(result)
```
vinyll: We probably want to display an actual HTML template rather than a json.

vinyll (changes requested):
I guess the code is basic enough to be offered as a first search engine.
We may consider it and only later migrate to a search based on mongodb.

A couple of small things to fix.
A template for displaying results is missing. There's no point of merging this code if it remains unused.
Maybe @fuentesloic may be interested into pairing with you to build a page of results? May you open an issue for that?

templates/search/list.html
```diff
@@ -16,61 +17,37 @@
...
   {% endif %}
-
+  
   {% if articles %}
-  <section>
+  <article>
     <h2>articles finding</h2>
     <ul>
       {% for article in articles %}
-        <li>
```
vinyll: choose between `section > ul > li` or `section > article`. I wouldn't put `section > ul > li > article`

templates/search/list.html
```diff
@@ -16,61 +17,37 @@
...
-        <li class=profile-intro>
-          <a href={{ user.slug }}>
-            <h2 name=username>{{ user.username }}</h2>
-            <figure>
-              <img class=avatar-img src={{ user.avatar }} width=100 height=100 onerror="this.src='{{ url_for('static', filename='images/avatar.png') }}'">
-            </figure>
-          </a>
-        </li>
+      <li class=profiles>
+        <a href='@{{ user.slug }}'>
```
vinyll: use `{{ url_for_user(user) }}`

templates/search/list.html
```diff
@@ -16,61 +17,37 @@
...
+          <h2 name=username>{{ user.username }}</h2>
+          <figure>
+            <img class=avatar-img src={{ user.avatar }} width=100 height=100 onerror="this.src='{{ url_for('static', filename='images/avatar.png') }}'">
+          </figure>
+        </a>
```
vinyll: It seems like we are having this block over and over.
It might be time to make a reusable [template macro](http://jinja.pocoo.org/docs/2.9/templates/#macros) (or a reusable webcomponent).

vinyll (changes requested):

templates/search/list.html
fuentesloic: I agree with you @vinyll but if we create a component or macro for `avatar` we should do that in an other PR

templates/blog/macros.html
```diff
@@ -1,6 +1,8 @@
 {% macro article_list(articles, with_author=true) %}
+  <ul>
   {% for article in articles %}
   {% set url = url_for_article(article) %}
+  <li>
```
vinyll: Why adding `ul > li` here? We do have a serie of articles already.

templates/search/list.html
```diff
@@ -0,0 +1,51 @@
...
+</header>
+
+<section>
+  {% if users %}
+  <article>
+    <h2>neomad finding</h2>
+    <ul>
+      {% for user in users %}
+      <li class=profiles>
+        <a href='@{{ user.slug }}'>
```
vinyll: use `{{ url_for_user(user) }}`

templates/search/list.html
```diff
@@ -0,0 +1,51 @@
...
+    </ul>
+  </article>
+  {% endif %}
+  
+  {% if articles %}
+  <article>
+    <h2>articles finding</h2>
+    <ul>
+      <li class=articles>
+        {{ article_list(articles=articles) }}
```
vinyll: `article_list` inside of one `li`? You will find the list of articles in one single `<li>`

templates/search/list.html
```diff
@@ -0,0 +1,51 @@
...
+  <form action='{{ url_for('search') }}' method=get>
+    <input id=search type=search name=q placeholder="search for neomads and articles">
+    <input type=submit value=search>
+  </form>
+</header>
+
+<section>
+  {% if users %}
+  <article>
+    <h2>neomad finding</h2>
```
vinyll: `{{users|length}} neomads found`

templates/search/list.html
```diff
@@ -0,0 +1,51 @@
...
+
+{% block extra_css %}
+<link rel=stylesheet href=//netdna.bootstrapcdn.com/font-awesome/4.1.0/css/font-awesome.css>
+<link rel=stylesheet href={{ url_for('static', filename='css/search.css') }}>
+<link rel=stylesheet href={{ url_for('static', filename='css/_article_list.css') }}>
+{% endblock %}
+
+{% block main %}
+<header>
+  <form action='{{ url_for('search') }}' method=get>
```
vinyll: If you use HTML quotes, use double quotes: `action="{{ url_for('search') }}"`

vinyll (changes requested):

templates/search/list.html
fuentesloic: done

templates/search/list.html
fuentesloic: done

templates/search/list.html
vinyll: ?

templates/search/list.html
fuentesloic: I did the same thing for `<h2>articles finding</h2>`, good idea 👍

vinyll: I guess so

templates/search/list.html
vinyll: Just push and it will automatically clear out these comments when they're fixed ;)

templates/blog/macros.html
fuentesloic: nop, `articles` was printing `<article>`, @Ydrasil fix that in the template and this morning I fix the css (but all of that are on the ydrasil/neomad repo), It's going to be good :)

templates/search/list.html
fuentesloic: oooh that was that ^^ thx

fuentesloic (approved):
For me it's ok

search/test.py
```diff
@@ -0,0 +1,76 @@
+import json
+from unittest import TestCase
+
+from search.views import *
```
vinyll: `from search import views`

search/test.py
```diff
@@ -0,0 +1,76 @@
...
+        user = User(email='john@test.com',
+                    username='john').set_password('search').save()
+        for i in range(0, 11):
+            User(email='fries{}@test.com'.format(i),
+                 username='friesuser{}'.format(i)).set_password(
+                        'search').save()
+            article = Article(title='<h1>Title</h1>',
+                              content='My fries are cold')
+            article.author = user
+            article.save()
```
vinyll: Simplification:
```python
Article(title='<h1>Title</h1>',
            content='My fries are cold',
            author=user).save()
```

search/test.py
```diff
@@ -0,0 +1,76 @@
...
+        response = self.client.get('/search?q=fries')
+        self.assertEqual(response.status_code, 200)
+        count = 0
+        for word in response.data.split():
```
vinyll: `response.data` [is deprecated](http://flask.pocoo.org/docs/0.12/api/#flask.Response.data)

search/test.py
```diff
@@ -0,0 +1,76 @@
...
+            if b'fries' in word:
+                count = count + 1
+        # /search?q=fries appear in response.data and
+        # href for users appear twice
+        self.assertEqual(count, 31)
```
vinyll: You may want to replace the whole previous block (line 70 > 76) by:
```python
self.assertEqual(request.get_data().count('fries'), 31)
```

vinyll: However, it's very unclear why testing twice 10 elements make 31 (and therefore unmaintainable in the longer term for the future you or other devs).
Maybe you want to make 3 tests: one for counting the 10 `href=…fries`, another for counting the 10 others items, again for the next 10, and again for the 1.

templates/blog/macros.html
vinyll: I'm still not confident we need `ul > li > article` and not just a serie of `article`, but if you have good reasons I'll rely on you.

templates/search/list.html
```diff
@@ -0,0 +1,51 @@
...
+    </ul>
+  </article>
+  {% endif %}
+  
+  {% if articles %}
+  <article>
+    <h2>{{articles|length}} articles found</h2>
+    <ul>
+      <li class=articles>
+        {{ article_list(articles=articles) }}
```
vinyll: `article_list` will be replaced by `<ul><li>…</li><li>…</li><li>…</li>…</ul>`

and here we insert this whole list into a `<ul><li>REPLACED LIST</li></ul>`

Therefore, we'll result with:

```html
<ul><li>
  <ul><li>…</li><li>…</li><li>…</li>…</ul>
</li></ul>
```
View the one single `li` in the first level that will contain a list of items.

In short, we don't want the `<ul><li>` at line 42, 43, 45, 46

vinyll (changes requested):

vinyll:
ok for me on the frontend side.
Still the backend to get fixed.

search/test.py
Ydrasil: Do we add this improvement in this commit ?

search/test.py
Ydrasil: The problem is that if we want to search one user, that is actually exist, and we have one result on the app. The name that we have search will appear 3 times on the data. 
- In the URL that is in the data
- In href
- In the `<h2>`

An example with 'emailtest' below :
```
[...]
<meta property=og:url content="http://0.0.0.0:5000/search?q=searchtest">
[...]
<li class=profiles>
        <a href='/@searchtest'>
          <figure>
            <img class=avatar-img src=/static/uploads/avatars/597b51536ce8df4c77d3ac61 width=100 height=100 onerror="this.src='/static/images/avatar.png'">
          </figure>
          <h2 name=username>searchtest</h2>
        </a>
      </li>
[...]
```

I'm agree that is not very maintainable :confused:

vinyll: That's why we should not just search for "searchtest" for checking that the value exists, but search for the value in a context: we should have 10 results similar to `<h2 name=username>searchtest</h2>`.

vinyll (approved):

PR #136, opened 2017-07-21
Check if the image still in the article when edit fix #135

### LINKED ISSUE #135: deleting an image in article don't delete it from the file system

Ydrasil:


vinyll (approved):

vinyll:
We'll need to simplify the logic of these 3 loops: https://github.com/Ydrasil/neomad.org/blob/82716b86aee89b9888bd1fec0e4b7f04e7230464/blog/models.py#L72

One should be enough

PR #140, opened 2017-07-27
Remove <br> from titles fix #139

### LINKED ISSUE #139: A title takes a `<br>` and is not cleaned
[Some tests](https://github.com/neomad-team/neomad.org/pull/131/files#diff-d5faad65f5d3bbc77dd08e78c4449fceR30) are done writing `<br>` in titles.
No HTML should be saved in a title when saving an article.

Ydrasil:


blog/test.py
```diff
@@ -93,10 +93,10 @@ def test_edit_article(self):
             'password': 'testtest',
         }
         self.client.post('/login', data=data, follow_redirects=True)
-        data = {'title': 'title', 'content': 'content'}
+        data = {'title': '<p>title<br></p>', 'content': 'content'}
```
vinyll: No need to add html here as you're not testing it. Just keep it miminum.

blog/test.py
```diff
@@ -93,10 +93,10 @@ def test_edit_article(self):
...
         self.client.post('/article/write', data=data)
         article = Article.objects.first()
-        data = {'title': 'title', 'content': 'another content'}
+        data = {'title': '<p>title<br></p>', 'content': 'another content'}
```
vinyll: same here

blog/test.py
```diff
@@ -117,9 +117,9 @@ def test_delete_an_article(self):
         self.assertEqual(result.status_code, 302)
 
     def test_title_clean_html(self):
-        article = Article(title='title',
+        article = Article(title='<p>title<br></p>',
                           content='<p>content</p>').save()
-        self.assertEqual(article.title, 'title')
+        self.assertEqual(article.title, '<p>title</p>')
```
vinyll: title should not have any html at all. It should just be a string.
The following test should pass:

```python
article = Article(title='<p>title<br></p>', content='<p>content</p>').save()
self.assertEqual(article.title, 'title')
```

vinyll (changes requested):

PR #143, opened 2017-08-07
Simplify blog image cleaning

Ydrasil:
@vinyll

### DIFF
--- a/blog/models.py
+++ b/blog/models.py
@@ -68,14 +68,10 @@ class Article(db.Document):
                 images.append(img_url)
             else:
                 images.append(data)
+        for outdated_image in set(self.images) - set(images):
+            os.remove(os.path.join(self.get_images_path(),
+                                   os.path.basename(outdated_image)))
         self.images = images
-        images_name = [os.path.basename(self.images[idx])
-                       for idx, image in enumerate(self.images)]
-        images_to_delete = [image for image in
-                            os.listdir(self.get_images_path())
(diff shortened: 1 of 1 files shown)


blog/models.py
```diff
@@ -69,12 +69,11 @@ def extract_images(self):
...
-                       for idx, image in enumerate(self.images)]
-        images_to_delete = [image for image in
-                            os.listdir(self.get_images_path())
-                            if image not in images_name]
-        for image in images_to_delete:
+        for image in [image for image in
+                      os.listdir(self.get_images_path())
+                      if image not in [os.path.basename(self.images[idx])
+                                       for idx, image in enumerate(
+                                       self.images)]]:
```
vinyll: This is a loop in a loop in a loop (n^3).

if you write this code before saving images in line 71 (therefore moving line 71 after this algorithm) you should retrieve the previous images with `self.images`.
Therefore you could do:

```python
for outdated_image in set(self.images) - set(images):
    os.remove(os.path.join(self.get_images_path(), outdated_image))
self.images = images
```
**untested**

vinyll (changes requested):

vinyll:
@Ydrasil: did you check it actually work with this new version of the code? Did you see the file deleted?

Ydrasil:
Yes It's tested, the file is correctly deleted

vinyll (approved):

Issue #120, opened 2017-06-27
bug with API spots


fuentesloic:
Hey, I spent two hours to finally find out ... we have a small bug with the API and that breake `/around/` in Dev :s

![capture d ecran 2017-06-27 a 17 45 06](https://user-images.githubusercontent.com/8755623/27611531-7ed9d1fe-5b60-11e7-90ee-81c46d6249bb.png)

So as you can see, some weird stuff appears: 

- `_id` is now an object with : `$oid` (it's the same for user and date with `$date`)

- `location` should be `location.latitude` & `location.longitude`

Ydrasil:
@fuentesloic I've try something, tell me if this json is ok to you or if I have to change something

![json_test_neomad_around](https://user-images.githubusercontent.com/10630464/27696164-3e007786-5cf1-11e7-95f1-e97f4d987d58.png)

vinyll:
LGTM. Just wondering if we should name `wifi` or `wifi_quality` and `power` or `power-available`.
If we ever have something else about the wifi, we want `wifi_quality`, if not `wifi` – shorter – is better.
@fuentesloic: the question is to you, what say you?

vinyll:
@Ydrasil note that @fuentesloic and myself has assigned ourselves as we had worked on it yesterday to fix it. We probably have the same final code as you do, unpushed yet ;)
We're also fixing the database structure which is not clean enough to make the API consistent (some empty comments are `None`, some are `[]`, some are `[""]`), we should have one way to represent a type of value (probably `[]`). Similar requirement for other fields.

fuentesloic:
@vinyll, @Ydrasil  I take an example with `wifi`
I prefer just `wifi` and if one day that evolve with for exemple `wifi-quality` and `wifi-available` we could just have something like:

```
wifi : 
      quality: 3
      passeword: superDuper_P@sseword
```

fuentesloic:
so @Ydrasil, you not push your modifications right ?
for the name `wifi` and `power` is good for me ! after if we are more detailled informations we could evolved as a real object like my previous comment

Ydrasil:
I haven't push my modification. As @vinyll have said it, we have probably the same code and I haven't fix the database structure.

vinyll:
@Ydrasil: It's always good to push work you've done; it gives ideas ;)

Ydrasil:
I've push it then #127

vinyll:
fixed

PR #145, opened 2017-08-10
cleaned api spots migration and preparing new version of API

vinyll:


Ydrasil (approved):

vinyll:
fixes #120

vinyll:
@Ydrasil: I've added [this commit](https://github.com/neomad-team/neomad.org/pull/145/commits/ab21b42a5a8e6e11599868719adcae63c6e1025a) after your review.
Can you please revisit?

Ydrasil (approved):
What do you about instead of creating `data` from `spot.to_mongo()`, create data by hand for avoiding delete some fields like `del data[_id]`

vinyll:
> What do you about instead of creating `data` from `spot.to_mongo()`, create data by hand for avoiding delete some fields like `del data[_id]`

That would be an option. However when using `to_mongo()`, any field that would be added to the model would automatically be present in the API too.
To be considered and discussed…

Issue #142, opened 2017-08-03
Homepage improvments


vinyll:
When the arrow is moving, the entire contents after it are moving along:

![aug-04-2017 00-28-48](https://user-images.githubusercontent.com/145172/28946635-ebf29d9e-78ab-11e7-925b-4b1c052634bf.gif)

---

The _become a nomad_ button has an unexpected behavior:
![aug-04-2017 00-31-33](https://user-images.githubusercontent.com/145172/28946718-504b11e0-78ac-11e7-83d7-09d885f4bf63.gif)

---

The console triggers the following JS error: `TypeError: avatar is null – utils.js:36:3`

---

Buttons padding in the second section could have a larger padding, like the "Read more" from section 3.

newick:
when you scroll (or get focus on the page, not sure) it scroll again after a delay

on the last FF version 

<img width="1437" alt="capture d ecran 2017-08-04 a 08 10 56" src="https://user-images.githubusercontent.com/657454/28958561-1c1366ac-78f7-11e7-9b84-3a6b6f9dc9f5.png">

Issue #4, opened 2017-03-22
Navigate to /profile for editing my profile


vinyll:
As a member, I can't go to edit my profile unless I know the URL (_/profile_)

vinyll:
Either we make a link to the /profile page from the upper bar for example, or if the user page is mine, we are in edit mode (just like the articles now)

vinyll:
it's now /@_myusername_

PR #146, opened 2017-08-12
correcting design from dev test

fuentesloic:


templates/blog/macros.html
```diff
@@ -2,7 +2,7 @@
   <ul>
   {% for article in articles %}
   {% set url = url_for_article(article) %}
-  <li class=articles>
+  <li class=article>
```
vinyll: I still wonder why you need a `ul > li` (with a class article!) to have an… article?
We did things like that when the `<article>` tag didn't exist

templates/blog/macros.html
```diff
@@ -21,7 +21,7 @@
     <a href={{ url }} lang={{ article.language }}>
       <h1>{{ article.title|title }}</h1>
       <figure>
-        <img src="{{ article.image | default('/static/images/articles/default.png', true) }}" width=100%>
+        <img src="{{ article.image | default('/static/images/article/default.png', true) }}" width=100%>
```
vinyll: bewarn that the _article**s**_ (with an "s") folder must exist and is used for uploaded contents.
It may be a better place to keep it in the same folder as uploaded.
The difference would be that _articles/default.png_ would be taken by default and _articles/1234/5678_ would be taken for image 5678 of article 1234.

vinyll (changes requested):

templates/blog/macros.html
fuentesloic: Ok I agree with you, I read a lot of article about it (yes I was thinking about it during my trip ^^).
I will change all the structure :)

vinyll:
Hey, there's a misunderstanding in commit names. you write "fix #4: …" which refers [to issue number 4](https://github.com/neomad-team/neomad.org/issues/4).
What you meant to do is something like "fix #142: homepage contents remain still with arrow moving".
When there's no issue open yet (like for `article*s*` semantic), you can't refer an issue. Thus you can just write "fix semantic moving article tags out of li".

Notice how Github is automatically creating a link in your commit names when you insert a `#`. ie: `fix #4` is automagically linking to issue number 4.

static/css/main.css
```diff
@@ -37,12 +37,12 @@ figure {
 }
 
 a {
-  color: #279e6aee;
+  color: rgba(39, 158, 106, .9);
```
vinyll: Note that this is not only for Chromium support, but also for Edge (latest IE – if we consider supporting this?).
[See the compatibility table](http://caniuse.com/#search=rgba)

PR #150, opened 2017-08-13
Privacy improve

vinyll:
Tests were working, but not the user interface.
Therefore users who were trying to delete their location did not react as expected (error 500).


vinyll:
Note that commit "detecting test env and setting config file accordingly" belongs to another PR but was necessary here (I had lost all my local user data running the tests).

core/__init__.py
```diff
@@ -1,12 +1,16 @@
...
+import sys
+from flask import Flask
 
 from . import mongoengine as db
 
 
-PROJECT_PATH = os.path.abspath(os.path.dirname(__file__) + '/..')
-os.environ.setdefault('SETTINGS', PROJECT_PATH + '/settings.py')
+settings_file = 'settings.py'
+if 'unittest' in sys.argv[0]:
```
Ydrasil: :100:

Ydrasil (approved):

PR #152, opened 2017-08-13
added location name for easy privacy locations editing

vinyll:
Guys, this is a mindblow! 🍸 

Users will have a list of city/country they visited for easier deletion!
I'm crazy about this as it means we're going to reuse this feature. To be continued 😈 

screen shot 2017-08-13 at 7 45 28 pm


Ydrasil:
I don't know why circleci fail here, maybe try to rebuild it

vinyll:
@Ydrasil: Still broken, even when reruning the tests

Ydrasil:
I'll try to search where does issues come from.

vinyll:
@Ydrasil: any clue so far? It would be great to get this merged.

vinyll:
It seems that the template is using `user.current_location`. Maybe wrapping it with an `{% if %}` block as you did on another PR would do the trick

fuentesloic:
nobody tried to wrap `user.current_location` ??

Issue #39, opened 2017-04-09
Search based on keywords


vinyll:
Once we have enough articles it may be interesting to be able to search through content and users

amberfrancis:
Can add WP-like tags based on location terms, topic, advice, personal etc

fuentesloic:
for now we have not enough content to search anything, after discussion (and user test) you have more chance to be desapointing after a search without result than use good keywords and find something.

We have to discuss more about this feature ... maybe is too early ...

My point of view for now is:

- we could decide to create some tags to help neomad writter to define the topic of their artciles,
- then we offer to reader users to filter by this same tags all the article to see that they want read according the topic wanted

tah-dahhhhh! :)

The result could be an `aside` in the article_list with all the tag, and on click we hidden all the article that don't have the tag.

What do you think about it ??

vinyll:
If it's decided we don't implement the _search_ code, we should not merge anything that is search related (backend and frontend). If it's merged, we should undo these commits from master.
Let's **avoid** making any new commit to remove what was done.

PR #154, opened 2017-08-14
WIP: move search bar into the header

fuentesloic:


static/js/search.js
```diff
@@ -0,0 +1,16 @@
+url = window.location.href
```
vinyll: You may be interested into the following magic:

```js
const query = window.location.search.slice(1).split('&').map(kv => kv.split('=')).map(arr => [arr[0], unescape(arr[1])])
const q = query.get('q')

Array.from(document.querySelectorAll('#search input')).forEach(input => {
  input.value = q
})
```

fuentesloic:
capture d ecran 2017-08-14 a 16 14 52-fullpage

vinyll:
I'd rather imagine something more discreet and incrusted, such as:

search example

@AymericFaivre / @newick: opinions or suggestions?

AymericFaivre:
Hey !

Yep, maybe just put the magnifying glass on the top left, and deploy the form on click. Maybe with just the glass in green. (like on medium, I like this kind of interaction)

Or put the first or second @vinyll's example to the top left and skin it by putting the magnifying glass in green and the border radius at max.

newick:
I agree with @AymericFaivre for the interaction and the rest :D

fuentesloic:
following the issue #39 
So for now, not enought content

vinyll:
According to https://github.com/neomad-team/neomad.org/issues/39#issuecomment-323245455, this PR is being postponed.

PR #159, opened 2017-08-23
Refactor test and templating condition

Ydrasil:
Add base abstract class for tests classes


blog/test.py
```diff
@@ -11,75 +12,52 @@
...
         self.assertIn(b'<article class=preview>', result.data)
 
     def test_read_article_unauthenticated(self):
-        article = Article(title='<h1>title</h1>', content='<p>content</p>')
-        article.author = self.user
-        article.save()
+        article = Article(title='title',
+                          content='<p>content</p>',
+                          author=self.create_user()).save()
         result = self.client.get('/@emailtest/{}-{}'.format(article.slug,
```
vinyll: This should end with a slash

blog/test.py
```diff
@@ -11,75 +12,52 @@
...
                                                             str(article.id)))
-        self.assertEqual(result.status_code, 301)
+        self.assertEqual(result.status_code, 200)
```
vinyll: That's a good point, 301 didn't have much sense.
Could you also test the article is displayed (something like `self.assertContains('<h1>title</h1>', result.data)`)

blog/test.py
```diff
@@ -11,75 +12,52 @@
...
     def test_read_article_authenticated(self):
-        data = {
-            'email': 'emailtest@test.com',
-            'password': 'testtest',
-        }
-        self.client.post('/login', data=data, follow_redirects=True)
-        article = Article(title='title', content='<p>content</p>')
-        article.author = self.user
-        article.save()
+        user = self.create_user()
```
vinyll: Why do you still need this now that you have `self.user`?

blog/test.py
```diff
@@ -11,75 +12,52 @@
...
+        self.login_user()
+        article = Article(title='title',
+                          content='<p>content</p>',
+                          author=user).save()
         result = self.client.get('/@emailtest/{}-{}'.format(article.slug,
```
vinyll: ending slash

blog/test.py
```diff
@@ -11,75 +12,52 @@
...
 
     def test_article_write_get(self):
-        data = {
-            'email': 'emailtest@test.com',
-            'password': 'testtest',
-        }
-        self.client.post('/login', data=data, follow_redirects=True)
+        self.create_user()
+        self.login_user()
         result = self.client.get('/article/write')
```
vinyll: Wondering why we don't have the ending slashes. Did you rebase from master?

blog/test.py
```diff
@@ -11,75 +12,52 @@
...
         self.assertEqual(result.status_code, 200)
 
     def test_article_write_post_empty_field(self):
-        data = {
-            'email': 'emailtest@test.com',
-            'password': 'testtest',
-        }
-        self.client.post('/login', data=data, follow_redirects=True)
+        self.create_user()
```
vinyll: `self.user`?

blog/test.py
```diff
@@ -11,75 +12,52 @@
...
         result = self.client.post('/article/write', data=data)
         self.assertEqual(result.status_code, 400)
 
     def test_article_write_post(self):
-        data = {
-            'email': 'emailtest@test.com',
-            'password': 'testtest',
-        }
-        self.client.post('/login', data=data, follow_redirects=True)
+        self.create_user()
```
vinyll: same

blog/test.py
```diff
@@ -88,11 +66,8 @@ def test_article_write_post(self):
         self.assertEqual(result.status_code, 201)
 
     def test_edit_article(self):
-        data = {
-            'email': 'emailtest@test.com',
-            'password': 'testtest',
-        }
-        self.client.post('/login', data=data, follow_redirects=True)
+        self.create_user()
```
vinyll: and here

blog/test.py
```diff
@@ -104,11 +79,8 @@ def test_edit_article(self):
         self.assertEqual(result.status_code, 302)
 
     def test_edit_article_with_empty_data(self):
-        data = {
-            'email': 'emailtest@test.com',
-            'password': 'testtest',
-        }
-        self.client.post('/login', data=data, follow_redirects=True)
+        self.create_user()
```
vinyll: reuse `self.user`

blog/test.py
```diff
@@ -118,11 +90,8 @@ def test_edit_article_with_empty_data(self):
         self.assertEqual(result.status_code, 400)
 
     def test_delete_an_article(self):
-        data = {
-            'email': 'emailtest@test.com',
-            'password': 'testtest',
-        }
-        self.client.post('/login', data=data, follow_redirects=True)
+        self.create_user()
```
vinyll: and again

core/test.py
```diff
@@ -1,8 +1,37 @@
...
 from unittest import TestCase
 
+from core import app
+from user.models import User
+from blog.models import Article
 from .utils import clean_html
 
 
+class BaseTest(TestCase):
+    __metaclass__ = ABCMeta
```
vinyll: funny how we discussed we never need to use metaclass.
What's the point of using metaclass here? 
If you remove this it will still work.

core/test.py
```diff
@@ -1,8 +1,37 @@
...
+        Article.objects.delete()
+
+    def login_user(self):
+        data = {
+            'email': 'emailtest@test.com',
+            'password': 'testtest',
+        }
+        self.client.post('/login', data=data, follow_redirects=True)
+
+    @staticmethod
```
vinyll: why making it static? you could do `self.create_user()` or even have `self.user` in he `setUp`

user/test.py
```diff
@@ -2,32 +2,15 @@
 from unittest import TestCase
 
 from core import app
+from core.test import BaseTest
```
vinyll: you don't need to import this, it's an abstract class that you only need when extending.

core/test.py
```diff
@@ -1,8 +1,37 @@
...
+    __metaclass__ = ABCMeta
+
+    def setUp(self):
+        self.client = app.test_client()
+
+    def tearDown(self):
+        User.objects.delete()
+        Article.objects.delete()
+
+    def login_user(self):
```
vinyll, line 18: You only want to do this for testing the `/login/` view with its process.

Otherwise you just login you user:

```python
from flask_login import login_user

login_user(self.user)
```

That's what is happening in the `/login/` view ;)

vinyll (changes requested):

core/test.py
Ydrasil: I've try to log an user with `flask_login.login_user` 
Firstly I had an issue : `RuntimeError: working outside of application context` 
So I've add `with app.test_request_context():`, even with `with app.request_context():` It has remove the `RuntimeError` but the user is still not authenticated.
I've also change the arguments of `login_user`, I'm stuck with this issue

blog/test.py
Ydrasil: I've try to put a `self.user` in the `setUp` but I've a `notUniqueError` when running tests

core/test.py
vinyll: I see your point. I don't have any cleaner workaround, so it's good for me to keep the `/login/` call for login in.

blog/test.py
vinyll: That is because you do a `self.user` and then you try to recreate the same user with the same data in the same test. Once `self.user` exists, you should check that you're not duplicating this user ;)

blog/test.py
```diff
@@ -11,75 +12,52 @@
...
         self.assertEqual(result.status_code, 200)
 
     def test_articles_with_an_article(self):
-        article = Article(title='<h1>title</h1>', content='<p>content</p>')
-        article.author = self.user
-        article.save()
+        Article(title='title',
+                content='<p>content</p>',
+                author=self.create_user()).save()
         result = self.client.get('/articles')
```
vinyll: ending slash

blog/test.py
```diff
@@ -11,75 +12,52 @@
...
         self.assertIn(b'<article class=preview>', result.data)
```
vinyll: test the title instead of some random html

blog/test.py
```diff
@@ -11,75 +12,52 @@
...
     def test_article_write_post(self):
-        data = {
-            'email': 'emailtest@test.com',
-            'password': 'testtest',
-        }
-        self.client.post('/login', data=data, follow_redirects=True)
+        self.create_user()
+        self.login_user()
         data = {'title': 'title', 'content': '<p>content</p>'}
         result = self.client.post('/article/write', data=data)
```
vinyll: ending slash

blog/test.py
```diff
@@ -104,11 +79,8 @@ def test_edit_article(self):
...
     def test_edit_article_with_empty_data(self):
-        data = {
-            'email': 'emailtest@test.com',
-            'password': 'testtest',
-        }
-        self.client.post('/login', data=data, follow_redirects=True)
+        self.create_user()
+        self.login_user()
         data = {'title': 'title', 'content': 'content'}
         self.client.post('/article/write', data=data)
```
vinyll: ending slash

blog/test.py
```diff
@@ -118,11 +90,8 @@ def test_edit_article_with_empty_data(self):
...
     def test_delete_an_article(self):
-        data = {
-            'email': 'emailtest@test.com',
-            'password': 'testtest',
-        }
-        self.client.post('/login', data=data, follow_redirects=True)
+        self.create_user()
+        self.login_user()
         data = {'title': 'title', 'content': 'content'}
         self.client.post('/article/write', data=data)
```
vinyll: ending slash

core/test.py
```diff
@@ -1,8 +1,37 @@
...
+        data = {
+            'email': 'emailtest@test.com',
+            'password': 'testtest',
+        }
+        self.client.post('/login', data=data, follow_redirects=True)
```
vinyll: ending slash

blog/test.py
```diff
@@ -11,72 +11,50 @@
...
 
     def test_articles_with_an_article(self):
-        article = Article(title='<h1>title</h1>', content='<p>content</p>')
-        user = self.create_user()
-        article.author = user
-        article.save()
+        user = self.user
+        Article(title='Title for an article',
+                content='<p>content</p>',
+                author=user).save()
```
vinyll: `author=self.user`, so you can get rid of the line above

blog/test.py
```diff
@@ -99,7 +76,6 @@ def test_edit_article(self):
         self.assertEqual(result.status_code, 302)
```
vinyll, line 76: you should also check where it redirects to

blog/test.py
```diff
@@ -86,7 +64,6 @@ def test_article_write_post(self):
         self.assertEqual(result.status_code, 201)
```
vinyll: Maybe you could check the article has effectively been saved in the db

blog/test.py
```diff
@@ -11,72 +11,50 @@
...
     def test_read_article_unauthenticated(self):
-        article = Article(title='title',
+        article = Article(title='title for an article',
                           content='<p>content</p>',
-                          author=self.user()).save()
+                          author=self.user).save()
         result = self.client.get('/@emailtest/{}-{}/'.format(article.slug,
-                                                            str(article.id)))
+                                                             str(article.id)))
+        self.assertIn(b'title for an article', result.data)
```
vinyll: check the status first, then the body.
The purpose of doing so is that if the status is wrong, the error is more explicit

blog/test.py
```diff
@@ -11,72 +11,50 @@
...
-        user = self.create_user()
         self.login_user()
-        article = Article(title='title',
+        article = Article(title='title for article',
                           content='<p>content</p>',
-                          author = user).save()
+                          author=self.user).save()
         result = self.client.get('/@emailtest/{}-{}/'.format(article.slug,
                                                              str(article.id)))
+        self.assertIn(b'title for article', result.data)
```
vinyll: check status first, then the body

core/test.py
```diff
@@ -1,8 +1,35 @@
 from unittest import TestCase
+from flask_login import login_user
+import mongoengine
```
vinyll: Why do you import `mongoengine`?

core/test.py
```diff
@@ -1,8 +1,35 @@
...
+    def tearDown(self):
+        User.objects.delete()
+        Article.objects.delete()
+
+    def login_user(self):
+        data = {
+            'email': self.user.email,
+            'password': 'testtest',
+        }
+        self.client.post('/login/', data=data, follow_redirects=True)
```
vinyll: why do you `follow_redirects` if you don't need the next page?

vinyll (changes requested):

blog/test.py
```diff
@@ -11,124 +11,91 @@
...
         article = Article.objects.first()
-        data = {'title': '<p>title<br></p>', 'content': 'another content'}
-        result = self.client.post('/article/{}/edit'.format(str(article.id)),
+        data = {'title': 'title', 'content': 'another content'}
+        result = self.client.post('/article/{}/edit/'.format(str(article.id)),
                                   data=data)
         article = Article.objects.first()
         self.assertEqual(article.content, '<p>another content</p>')
         self.assertEqual(result.status_code, 302)
+        self.assertIn('/@{}/{}-{}/'.format(self.user.slug, article.title,
```
vinyll: Why `assertIn` instead `assertEqual`?

vinyll (commented):
Only need to compile the commits into a couple of them and we should be good 👍

Ydrasil:
@vinyll Can you rebase `neomad-team:privacy-location-names` with `master` ? Maybe it will explain the 19 commits

fuentesloic:
@vinyll you should check this branch and merge it (on master) because all the test are fixed here

PR #162, opened 2017-08-23
Leaflet migration

fuentesloic:


static/js/utils.js
```diff
@@ -1,5 +1,8 @@
 function alert(type, message, delay) {
   console.info(type, message)
+  if(!message) {
```
vinyll: mhhh… why would you call alert with only 1 argument?
What would be the type?

static/js/utils.js
```diff
@@ -1,5 +1,8 @@
...
+    message = type
+  }
   let notification = document.querySelector('#notification')
```
vinyll: `const notification = document.querySelector('#notification') || document.createElement('div')`
But why would `#notification` not be found? Isn't it supposed to be in the layout?
Notification is not just a plain blank `div`, it has attributes.

vinyll:
And what is this commit about changing alert and notifications doing in leaflet migration??

static/js/utils.js
fuentesloic: because when an user block the location from his browser, the browser run alert as `alert(message)`
so message is `undefined` and type is a message

static/js/utils.js
fuentesloic: you write this ... :)

fuentesloic:
It's not false :) I put it here because the probleme is appear when the browser send me `alert()` 
error during localization

static/js/utils.js
vinyll: Well, you let this go through the review without giving an explicit reason? 😛 
It's either not applicable anymore, or it's missing a comment to explain why I needed to do this :/

static/js/utils.js
vinyll: Are you saying that the browser is calling the `alert()` function by itsself when rejecting geolocation?
It would then make sense, though probably in another PR (as it's not specific to this PR)

vinyll:
> It's not false :) I put it here because the probleme is appear when the browser send me alert()
error during localization

Ok, I see the point.
Notice that the type would still exist and is used for the _classname_. for an `alert('test me now')`, you would end with a `<div class="test me now">test me now</div>`. We probably want to set a class by default (therefore a `type` which may default to "_info_") so the html does not break.

static/js/utils.js
fuentesloic: I will cut and paste your comments in the new PR

static/css/around-map.css
```diff
@@ -295,63 +295,34 @@ footer {
...
-}
-
-.current-marker {
-  border: 0.4rem solid #3ba77c !important;
-  z-index: 1;
-}
-
 .visible {
-  display: flex !important;
+  display: flex!important;
```
vinyll: why removing this space?

static/css/around-map.css
```diff
@@ -295,63 +295,34 @@ footer {
...
-  cursor: pointer;
-  width: 2rem;
-  height: 3rem;
-  border-radius: 0;
-  background-color: transparent;
-  border: transparent;
-  z-index: 2;
-  left: -0.4rem;
-  top: -2rem;
+.first-card {
```
vinyll: jsut curious, but why setting a class instead of using `:first-of-type` or `:first-child`

static/css/around-map.css
```diff
@@ -295,63 +295,34 @@ footer {
...
-  z-index: 2;
-  top: -1.9rem;
+.marker {
+  fill: #297ddb;
 }
-
-.mapboxgl-popup-content p {
-  text-align: center;
+.current-marker {
+  fill: #9B9B9B;
```
vinyll: lowercase?

static/css/around-map.css
```diff
@@ -295,63 +295,34 @@ footer {
...
+  /* need important to overide leaflet */
+  z-index: 9999!important;
```
vinyll: leaflet goes up to 1010 if I remember well. You may want to set it to that level instead.

static/css/main.css
```diff
@@ -91,7 +91,7 @@ body > header nav a {
 .avatar {
   cursor: pointer;
 }
-.avatar img {
+.avatar-img {
```
vinyll: why setting a class for the img?

static/css/trips-map.css
```diff
@@ -1,24 +1,20 @@
-h1 {
-  margin-bottom: 3rem;
-}
-
 #map {
   width: 100%;
-  min-height: 500px;
+  min-height: 62rem;
+  top: -3.5rem;
+  z-index: 0;
```
vinyll: `0`? That's the browser's default

static/css/trips-map.css
```diff
@@ -1,24 +1,20 @@
...
+}
+
+svg {
+  fill: #279E6A;
```
vinyll: lowercase

static/js/around-map.js
```diff
@@ -1,50 +1,70 @@
...
+// POIS creation
 const worker = new Worker('/static/js/webworker-around.js')
 let pois = []
 
 function addPoi (poi) {
-  const el = document.createElement('div')
-  el.classList.add('marker')
-  el.id = poi.id
+  const marker = L.marker(poi.location, {icon: icon, alt: poi.name}).addTo(map)
+  marker._icon.setAttribute('id', poi.id)
```
vinyll: Is it the only way to access the icon? Most of the time, a property that starts with a `_` should not be call as is. Its "private" hence hackish. No big deal though.

static/js/trips-map.js
```diff
@@ -1,81 +1,89 @@
...
+  iconSize: [30, 50],
+  iconAnchor: [15, 42],
+  popupAnchor: [0, -40],
+  viewBox: '300 100 1200 1200',
+  path: 'm 768,896 q 0,106 -75,181 -75,75 -181,75 -106,0 -181,-75 -75,-75 -75,-181 0,-106 75,-181 75,-75 181,-75 106,0 181,75 75,75 75,181 z m 256,0 q 0,-109 -33, -179 L 627,-57 q -16,-33 -47.5,-52 -31.5,-19 -67.5,-19 -36,0 -67.5,19 Q 413,-90 398, -57 L 33,717 Q 0,787 0,896 q 0,212 150,362 150,150 362,150 212,0 362,-150 150,-150 150,-362 z',
+  transform: 'matrix(1,0,0,-1,364.47458,1270.2373)'
+}
 
-  lineCoords.push(point.position)
+class Icon extends L.Icon {
```
vinyll: seems like copy and paste. We'll need to refactor this code into one single spot

static/js/trips-map.js
```diff
@@ -1,81 +1,89 @@
...
+  createIcon() {
+    const marker = document.createElement('div')
+    const options = this.options
+    marker.innerHTML = `<svg xmlns:cc="http://creativecommons.org/ns" xmlns:svg="http://www.w3.org/2000/svg" version="1.1"
+      viewBox="${options.viewBox}"
+      width="${options.iconSize[0]}px"
+      height="${options.iconSize[1]}px">
+      <g transform="${options.transform}"><path d="${options.path}"/></g>
+    </svg>`
+    this._setIconStyles(marker, 'icon')
```
vinyll: is there no method that doesn't start with a `_`?

static/js/trips-map.js
```diff
@@ -1,81 +1,89 @@
...
 
-  new mapboxgl.Marker(el, {offset:[0, -30]})
-      .setLngLat(current_location.reverse())
-      .setPopup(popup)
-      .addTo(map)
-}
+lineCoords.push(lastPosition.position)
+
+// Previous trips
+icon.options.className = "previous-marker"
```
vinyll: single quotes

templates/around/map.html
```diff
@@ -3,10 +3,14 @@
 {% block title %}Around you{% endblock %}
 
 {% block extra_css %}
-  <link rel=stylesheet href=https://api.mapbox.com/mapbox-gl-js/v0.34.0/mapbox-gl.css >
-  <link rel=stylesheet href={{ url_for('static', filename='css/trips-map.css') }}>
   <link rel=stylesheet href={{ url_for('static', filename='css/around-map.css') }}>
-  <script src=https://api.mapbox.com/mapbox-gl-js/v0.34.0/mapbox-gl.js></script>
+  <!-- leaflet -->
+  <link rel="stylesheet" href="https://unpkg.com/leaflet@1.2.0/dist/leaflet.css" integrity="sha512-M2wvCLH6DSRazYeZRIm1JnYyh22purTM+FDB5CsyxtQJYeKq83arPe5wgbNmcFXGqiSH2XR8dT/fJISVA1r/zQ==" crossorigin=""/>
```
vinyll: you very obviously copy and pasted. Code I've never seen before 😮 
You may remove quotes where not required ;)

templates/trips/map.html
```diff
@@ -9,23 +9,18 @@
 {% endblock %}
 
 {% block extra_css %}
-<link href=https://api.mapbox.com/mapbox-gl-js/v0.34.0/mapbox-gl.css rel=stylesheet>
 <link rel=stylesheet href={{ url_for('static', filename='css/trips-map.css') }}>
-<script src=https://api.mapbox.com/mapbox-gl-js/v0.34.0/mapbox-gl.js></script>
+<!-- leaflet -->
+<link rel="stylesheet" href="https://unpkg.com/leaflet@1.2.0/dist/leaflet.css" integrity="sha512-M2wvCLH6DSRazYeZRIm1JnYyh22purTM+FDB5CsyxtQJYeKq83arPe5wgbNmcFXGqiSH2XR8dT/fJISVA1r/zQ==" crossorigin=""/>
```
vinyll: same

vinyll (changes requested):

static/js/around-map.js
fuentesloic: I try many many many way without suceed.
the smart should be: `L.marker(poi.location, {icon: icon, id: poi.id})` that work for alt but not for the `id` and I dont know why :s

fuentesloic:
@vinyll could you check this code? for me it's done (for this part)
I refacto the creation map and marker cache (I hope you was thinking to do as I did ^^)

@AymericFaivre @newick : what do you think about the map and color marker ?
Like all know it, I'm not good with this kind of things, we have some suggestion from: 
[mapbox](https://www.mapbox.com/maps/#ready-to-use)
feel you free to propose other source if you want :)
let me know (help me plzzzzz ^^) wich one is better according your designer soul :)

For the marker I have 3 colors:
- blue for a standard marker,
- orange (eurk) when you hover a card (that change the color of its pair marker)
- green when you select the pair poi marker/card

thinks guyss :)

static/js/around-map.js
vinyll: Maybe `id` is a reserved key name

static/components/PoiCard.js
```diff
@@ -14,7 +14,7 @@ class PoiCard extends React.Component {
     }
     // cf: https://facebook.github.io/react/docs/handling-events.html
     this.hoverCard = this.hoverCard.bind(this)
-    this.clickCard = this.clickCard.bind(this)
+    // this.clickCard = this.clickCard.bind(this)
```
vinyll: remove the line if it's unused

static/components/PoiCard.js
```diff
@@ -57,14 +57,14 @@ class PoiCard extends React.Component {
...
-    urlFor(this.props.details.id)
-  }
+  // clickCard() {
+  //   let zoomLevel = 11
+  //   if(this.props.details.id === getHash() && map.getZoom() < 14) {
+  //     zoomLevel = 14
+  //   }
+  //   moveTo(this.props.details.location, zoomLevel)
+  //   urlFor(this.props.details.id)
+  // }
```
vinyll: remove if unused

static/css/around-map.css
```diff
@@ -296,28 +296,30 @@ footer {
...
-  border: 0.2rem solid #3ba77c;
-}
+/* marker */
 
 .marker {
   fill: #297ddb;
 }
+
+/* current marker/card hovering */
+.current-card h2{
```
vinyll: space before `{`
Where is the _hovering_ that the comment says?

static/js/around-map.js
```diff
@@ -123,24 +123,20 @@ function findPoi (id) {
...
+  return twins = [
+    { class: 'marker', div: marker },
+    { class: 'card', div: card }
+  ]
+}
+
+function highlight (poi_id) {
+  getTwins(poi_id).forEach(el => {
+    el.div.classList.toggle(`current-${el.class}`)
+  })
```
vinyll: I guess I'd replace the whole `getTwins` function and the forEach function with the simpler:
```js
document.getElementById(poi_id).classList.toggle('current-marker')
document.querySelector(`#card-${poi_id}`).classList.toggle('current-card')
```

vinyll (changes requested):

static/components/PoiCard.js
```diff
@@ -45,7 +45,7 @@ class PoiCard extends React.Component {
   }
 
   renderDistance(distance) {
-    if(!distance) return ''
+    if(distance < 15) return 'close to you'
```
vinyll: Why not taking a value like 100m?

static/css/around-map.css
```diff
@@ -311,13 +310,22 @@ footer {
   color: #9b9b9b;
 }
 .current-marker {
-  /* need 201 to overide leaflet native at 200  */
+  /* need 201 with important to overide leaflet native at 200  */
   z-index: 201 !important;
```
vinyll: Are you sure you're not supposed to go up to 1010? (leaflet's top layer)

static/js/around-map.js
```diff
@@ -84,9 +84,10 @@ window.onhashchange = _ => {
 }
 
 map.on('click', event => {
-  const poi = findPoi(event.originalEvent.target.id)
+  const poi = findPoi(event.originalEvent.target.ownerDocument.activeElement.id)
+  console.log(poi);
```
vinyll: remove

static/css/around-map.css
```diff
@@ -280,59 +290,43 @@ footer {
...
 
 /* marker */
-
 .marker {
   fill: #297ddb;
 }
+.hash-marker {
+  fill: #3ba77c;
+  /* leaflet native max at 1010 */
+  z-index: 1010!important;
```
vinyll: space before `!important` (for consistency)

static/css/_article_list.css
```diff
@@ -45,6 +45,13 @@
   display: block;
   margin-bottom: 0.2rem;
 }
+.meta a img {
+  display: inline-block;
+  vertical-align: middle;
+  border: 1px solid #666;
```
vinyll, line 51: you 😈 `666`

static/js/map.js
```diff
@@ -0,0 +1,45 @@
...
+  iconSize: [30, 50],
+  iconAnchor: [15, 42],
+  popupAnchor: [0, -40],
+  className: 'marker',
+  viewBox: '300 100 1200 1200',
+  path: 'm 768,896 q 0,106 -75,181 -75,75 -181,75 -106,0 -181,-75 -75,-75 -75,-181 0,-106 75,-181 75,-75 181,-75 106,0 181,75 75,75 75,181 z m 256,0 q 0,-109 -33, -179 L 627,-57 q -16,-33 -47.5,-52 -31.5,-19 -67.5,-19 -36,0 -67.5,19 Q 413,-90 398, -57 L 33,717 Q 0,787 0,896 q 0,212 150,362 150,150 362,150 212,0 362,-150 150,-150 150,-362 z',
+  transform: 'matrix(1,0,0,-1,364.47458,1270.2373)'
+}
+
+class Icon extends L.Icon {
```
vinyll, line 30: 👍

vinyll (approved):

templates/around/map.html
```diff
@@ -5,8 +5,8 @@
 {% block extra_css %}
   <link rel=stylesheet href={{ url_for('static', filename='css/around-map.css') }}>
   <!-- leaflet -->
-  <link rel=stylesheet href=https://unpkg.com/leaflet@1.2.0/dist/leaflet.css integrity=sha512-M2wvCLH6DSRazYeZRIm1JnYyh22purTM+FDB5CsyxtQJYeKq83arPe5wgbNmcFXGqiSH2XR8dT/fJISVA1r/zQ==/>
-  <script src="https://unpkg.com/leaflet@1.2.0/dist/leaflet.js" integrity="sha512-lInM/apFSqyy1o6s89K4iQUKg6ppXEgsVxT35HbzUupEVRh2Eu9Wdl4tHj7dZO0s1uvplcYGmt3498TtHq+log=="></script>
+  <link rel=stylesheet href=https://unpkg.com/leaflet@1.2.0/dist/leaflet.css integrity=sha512-M2wvCLH6DSRazYeZRIm1JnYyh22purTM+FDB5CsyxtQJYeKq83arPe5wgbNmcFXGqiSH2XR8dT/fJISVA1r/zQ== crossorigin=/>
```
vinyll: `crossorigin=`?
I'm not sure what this does, however it should have quotes `"` for it's empty value, or have no `=` at all!

vinyll (changes requested):

static/components/Comments.js
```diff
@@ -15,12 +15,11 @@ class Comments extends React.Component {
...
-      .map(key =>
-        <li><p>{this.state.comments[key]}</p></li>
-      )
-    // webnomade API have two cases for empty comments - issue #96
-    if(comments.length === 0 || comments[0] == [""]) {
+    .keys(this.state.comments)
+    .map(key =>
+      <li><p>{this.state.comments[key]}</p></li>
+    )
+    if(comments[0] == [""]) {
```
vinyll: single quotes 😛

vinyll (changes requested):

static/components/Comments.js
fuentesloic: crap! :)

templates/around/map.html
fuentesloic: without `=` we have an error

static/css/_article_list.css
fuentesloic: mouahahahahah

templates/around/map.html
vinyll: In which case you could then say `crossorigin=""`.
What's the error message BTW?

newick:
@fuentesloic could we have some screenshots ? :D

fuentesloic:
- blue for a standard marker,
- orange (eurk) when you hover a card (that change the color of its pair marker)
- green when you select the pair poi marker/card

output_jzx4gg

newick:
First of all, add a white border on every card, like that it won't move on hover.
Remember that hover don't work on mobile ;)

I'm ok with the idea, for the color, let's see with @AymericFaivre 😛

fuentesloic:
@newick thinks for it, concerning the mobile version there are not cards, just the map with a popup displayed when the user click on them

AymericFaivre:
Hey !
When you hover a card, maybe you could put all the none selected/hover pin in grey and keep the over pin in blue?

So the 2 pins (selected and hover) would be well localised. 

Is it possible? Or a good idea?

vinyll:
All good for me. I let you deal with the gfx guyz. @newick / @AymericFaivre: I'll let you approve this PR when you agree with @fuentesloic 😗

fuentesloic:
@AymericFaivre : yes sure it's possible, should we modify  the changement on card according the actions ?

vinyll:
Note that clicking twice on the same icon keeps it orange.
I would accuse `toggle`…

screen shot 2017-09-05 at 9 46 16 am

AymericFaivre:
@fuentesloic yep ;)
If it's possible to have a screenshot to be sure it's a good idea it would be perfect :)

fuentesloic:
@AymericFaivre : what do you think ?
output_gjqhvl

fuentesloic:
@vinyll that was the toggle ^^

static/css/around-map.css
```diff
@@ -44,6 +44,7 @@ main {
   background-color: #fefefe;
   cursor: pointer;
   margin-top: 2rem;
+  border: 0.2rem solid #fefefe;
 }
 .card:hover {
   box-shadow: 0 1rem 2rem 0 rgba(0,0,0,0.2);
```
vinyll: spaces after commas :p

static/css/around-map.css
```diff
@@ -323,19 +324,25 @@ footer {
...
+  fill: #8f8f8f;
+  /* leaflet native max at 1010 */
+  z-index: 1010 !important;
 }
 .current-marker {
-  fill: #fc8428;
-  /* current marker/card hovering with 1011 */
-  z-index: 1011 !important;
+  fill: #297ddb;
+  /* hash marker hovering with 1012 */
```
vinyll: could you rephrase this comment?

templates/trips/map.html
```diff
@@ -15,9 +15,8 @@
 <script src=https://unpkg.com/leaflet@1.2.0/dist/leaflet.js integrity=sha512-lInM/apFSqyy1o6s89K4iQUKg6ppXEgsVxT35HbzUupEVRh2Eu9Wdl4tHj7dZO0s1uvplcYGmt3498TtHq+log== crossorigin=></script>
 <!-- leaflet.leaflet-locatecontrol -->
 <link rel=stylesheet href=//maxcdn.bootstrapcdn.com/font-awesome/4.5.0/css/font-awesome.min.css>
-<link rel=stylesheet href=https://domoritz.github.io/leaflet-locatecontrol/dist/L.Control.Locate.min.css/>
 <script src=https://domoritz.github.io/leaflet-locatecontrol/dist/L.Control.Locate.min.js charset="utf-8"></script>
```
vinyll: no quotes

templates/around/map.html
```diff
@@ -3,14 +3,13 @@
...
-  <link rel=stylesheet href=//maxcdn.bootstrapcdn.com/font-awesome/4.5.0/css/font-awesome.min.css>
-  <link rel=stylesheet href=https://domoritz.github.io/leaflet-locatecontrol/dist/L.Control.Locate.min.css/>
-  <script src=https://domoritz.github.io/leaflet-locatecontrol/dist/L.Control.Locate.min.js charset="utf-8"></script>
+<link rel=stylesheet href={{ url_for('static', filename='css/around-map.css') }}>
+<!-- leaflet -->
+<link rel=stylesheet href=https://unpkg.com/leaflet@1.2.0/dist/leaflet.css integrity=sha512-M2wvCLH6DSRazYeZRIm1JnYyh22purTM+FDB5CsyxtQJYeKq83arPe5wgbNmcFXGqiSH2XR8dT/fJISVA1r/zQ== crossorigin=""/>
+<script src=https://unpkg.com/leaflet@1.2.0/dist/leaflet.js integrity=sha512-lInM/apFSqyy1o6s89K4iQUKg6ppXEgsVxT35HbzUupEVRh2Eu9Wdl4tHj7dZO0s1uvplcYGmt3498TtHq+log== crossorigin=""></script>
+<!-- leaflet.leaflet-locatecontrol -->
+<link rel=stylesheet href=//maxcdn.bootstrapcdn.com/font-awesome/4.5.0/css/font-awesome.min.css>
+<script src=https://domoritz.github.io/leaflet-locatecontrol/dist/L.Control.Locate.min.js charset="utf-8"></script>
```
vinyll: no quotes

vinyll:
`toggle` is here again!
clicking on a gray marker makes it green
clicking on a green marker makes it blue
clicking outside of a green marker removes the green one to select the one clicked and make it green.
You end with a bunch of grey markers, some blue markers, and one green.
Nice rainbow. What's the meaning of the blue?

screen shot 2017-09-05 at 7 13 18 pm

AymericFaivre:
I thinks it's good to me @fuentesloic , if we can remove the bug that @vinyll saw ;)

fuentesloic:
@vinyll It's perfectly works on my laptop, could you re build the preact app plz

fuentesloic:
@vinyll : I just forget to change the click action ...
Edit: worse than that I myself create the bug ^^ It's solve with the next commit

static/css/around-map.css
fuentesloic: I have to install a plugin to see this fucking invisible spaces (if you have a name ...)

vinyll: I just meant `box-shadow: 0 1rem 2rem 0 rgba(0, 0, 0, 0.2);` ;)

fuentesloic:
I think all is good now, another try required ;)

AymericFaivre:
Thanks !

vinyll (approved):
Just need to squash most commits to have only a few

PR #166, opened 2017-09-02
Home - Display last neomads

fuentesloic:
In the end of home page.

We displayed the 12 (arbitrarely maximum) last neomads to show to a visitor how cool looks the Neomads communauty (like you can see bolowing) 
capture d ecran 2017-09-02 a 18 22 18-fullpage
It's too many (too much) picture of me, but that works ^^

### DIFF
--- a/core/views.py
+++ b/core/views.py
@@ -11,4 +11,5 @@ from .helpers import url_for_user
 @app.route('/')
 def home():
     articles = Article.objects.all()[:3]
-    return render_template('home.html', articles=articles)
+    users = User.objects.all()[:12]
+    return render_template('home.html', articles=articles, users=users)
--- a/static/css/edit.css
+++ b/static/css/edit.css
@@ -71,7 +71,7 @@ form[edit] .toolbar .delete:hover {
   position: relative;
   padding-left: 6rem;
   cursor: pointer;
-  padding-top: .25rem;
+  padding-top: .5rem;
   z-index: 5;
 
 }
--- a/static/css/edit.css
+++ b/static/css/edit.css
@@ -71,7 +71,7 @@ form[edit] .toolbar .delete:hover {
   position: relative;
   padding-left: 6rem;
   cursor: pointer;
-  padding-top: .5rem;
+  padding-top: .25rem;
   z-index: 5;
 
 }
--- a/static/css/home.css
+++ b/static/css/home.css
@@ -76,19 +76,18 @@ section:first-child > div > h2 {
 
 #features,
 .articles,
+.profiles,
 footer {
   position: relative;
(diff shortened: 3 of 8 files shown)


newick:
1st question: do we have 12 users right now? :P
2nd question: do we need this kind of info on the homepage? It could be a dedicated page once you're logged. 

We need to have a setting to be hidden from this list.

vinyll:
I had been thinking about this feature and couln't figure out what place would be appropriate to display that list.

> like you can see bolowing

A mix between "below" and "in the following"? ❤️

fuentesloic:
@newick , @vinyll :
- I don't know how many neomads we have but It' work with less than 12 (we could change this qtty if you want)
- I agree with you with the privacy setting! it's mandatory (I will try to do it but ... it's python ^^)
- For the needs about having it or not, (for me) it's interesting to introduce real humans to show one of our principal values: we are a community, not just a database of content for nomad.
What do you think ?

@vinyll :
- I thought I was doing the right thing :)

vinyll:
@newick

> We need to have a setting to be hidden from this list.

I thought of 2 browsing modes:
- _private mode_: everything that may be a privacy concern (trips, around me, I recently subscribed) is accessible only by the user.
- _community mode_: geolocation is required, trips are saved, people around can see the user.

A user could toggle a mode or another at any moment from the interface.

Note that even in _private mode_, we may want to enable geolocation blurring for a user, which might be a privacy setting.

fuentesloic:
you could want be with not a public profile but use geolocalisation for your trips and around no ?

vinyll:
> you could want be with not a public profile but use geolocalisation for your trips and around no ?

Probably. I we were going down that road, we should decide which feature should be in which case.
The goal would be to implement both cases with **simplicity** in mind. No middle case for now.
exemple:
- private: geolocation is optional, around does not show people (you cannot see if you can't be seen), you cannot visit people's trips.
- community: geolocation is required. You can access everything.

This is just a suggestion. To be discussed ;0

fuentesloic:
Yes we have to discuss about that, because for me even if you decided to use the geoloc for you trips, around and read article BUT you also decided to be hidden by the others neomad I would like that this user could access to all the content.
I don't want punish him hiddening content to him because I decided to stay in no-public mode, If he want share, he share, if he don't want feel him free to naviguate as all the other users.

newick:
ok let's try with list on homepage and we'll change after if it's not good.

For privacy, should we discuss that on a dedicated issue ?

newick:
ok let's try with list on homepage and we'll change after if it's not good.

For privacy, should we discuss that on a dedicated issue ?

PR #167, opened 2017-09-03
fix animation to signup call to action

fuentesloic:
@vinyll it's following an issue that to find in the issue list


fuentesloic:
output_o308st

vinyll (approved):
👍   👏  👙 🏖

PR #169, opened 2017-09-04
User sorting

vinyll:
Latest created user appears first in the home list


Ydrasil (commented):
I'm wondering why there is this commit: `Around - form to add poi (#108)`
However the code seems correct to me

vinyll:
That's a good point @Ydrasil.
I believe I'm disrupted from Github's repo. I rebuilt it all to be fully synced.
Pushing again for an easier review ;)

vinyll:
Looks better now 👍

Ydrasil (approved):

PR #170, opened 2017-09-04
Error pages

fuentesloic:


static/css/error.css
```diff
@@ -0,0 +1,33 @@
+main {
+  height: 100%;
+  background: url('/static/error.jpg');
```
vinyll: you may want to move this to ../images/error.jpg

templates/error.html
```diff
@@ -0,0 +1,25 @@
+{% extends 'layout.html' %}
+
+{% block title %}{{ code }} - Digital Nomad Community{% endblock %} 
+{% block description %}Sorry for the inconveniance, let we bring back you to home.{% endblock %}
```
vinyll: inconvenience…
…let us bring you back home.

templates/error.html
```diff
@@ -0,0 +1,25 @@
...
+
+{% block extra_css %}
+<link rel=stylesheet href={{ url_for('static', filename='css/error.css') }}>
+{% endblock %}
+
+{% block main %}
+
+<section>
+  <h2>Dont' worry, it's just a {{ code }}</h2>
```
vinyll: Don't

templates/error.html
```diff
@@ -0,0 +1,25 @@
...
+  <a href=/ class=action-button>Go back to home</a>
+</section>
+
+{% endblock %}
+
+{% block extra_js %}
+<script>
+  console.debug('{{ error }}')
```
vinyll: This can probably be skipped

vinyll (commented):
hahaha

static/css/error.css
```diff
@@ -1,6 +1,6 @@
 main {
   height: 100%;
-  background: url('/static/error.jpg');
+  background: url('/static/images/error.jpg');
```
vinyll: a relative path `../static/images/error.jpg` would be less context dependant.

fuentesloic:
still looking for flask handler error ... I find that [flask error handler](http://flask.pocoo.org/docs/0.12/errorhandling/#error-handlers)
but not sure to catch the importance and side effect (for test or some stuff like that)

core/views.py
```diff
@@ -13,3 +13,12 @@ def home():
     articles = Article.objects.all()[:3]
     users = User.objects.all()[:3]
     return render_template('home.html', articles=articles, users=users)
+
+@app.errorhandler(404)
+def page_not_found(e):
+    return render_template('error.html', error=e, code=404), 404
+
+@app.errorhandler(500)
+def page_not_found(e):
```
vinyll: don't give twice the same name to functions.
`def server_error`, or `def error_500` for example

core/views.py
```diff
@@ -13,3 +13,12 @@ def home():
     articles = Article.objects.all()[:3]
     users = User.objects.all()[:3]
     return render_template('home.html', articles=articles, users=users)
+
+@app.errorhandler(404)
```
vinyll: - the list of possible errors: http://werkzeug.pocoo.org/docs/0.12/exceptions/

You may therefore be interested into `werkzeug.exceptions.NotFound` and `werkzeug.exceptions.InternalServerError`:

```python
from werkzeug.exceptions import NotFound, InternalServerError


@app.errorhandler(NotFound)
def error_404(e):
    return render_template('error.html', error=e, code=404), 404


@app.errorhandler(InternalServerError)
def error_500(e):
    return render_template('error.html', error=e, code=500), 500
```

However I'm not sure about the `,404` and `, 500` in the end of returns, or the the errorhandler would do it implicitely

fuentesloic: I'll try ... I was close ^^

vinyll (approved):

PR #178, opened 2017-09-10
fix bug from merge #115

fuentesloic:


vinyll (approved):
A new Pythonist is born 🐣 

👍

PR #182, opened 2017-09-18
TEST: Vuejs import styleguide - test with x-form login

fuentesloic:
capture d ecran 2017-09-18 a 12 18 46-fullpage

### DIFF
--- a/templates/auth/login.html
+++ b/templates/auth/login.html
@@ -8,78 +8,45 @@
 {% endblock %}
 
 {% block main %}
+
+<!-- it's a preview, css just here to fix css in styleguide -->
 <style>
-form {
-  max-width: 400px;
-  margin: 0 auto;
-  padding: 20px;
-  border: 1px solid #ccc;
-  background-color: white;
+main {
+  height: 100%;
 }
-label {
-  display: inline-block;
-  width: 100px;
-  margin-bottom: 20px;
-  font-weight: bold;
+#body {
+  display: flex;
+  justify-content: center;
+  height: 100%;
+  align-items: center;
 }
-input {
-  font: inherit;
-  padding: 0 5px;
-  border: 1px solid #ccc;
-  border-radius: 5px;
+form.login[data-v-6e1bd991] {
+  max-width: 35rem;
 }
-form button {
-  font: inherit;
-  font-weight: inherit;
-  background-color: #297ddb;
-  color: white;
-  border: 1px solid #6c6969;
-  border-radius: 7px;
-  padding: 5px 50px;
-  font-weight: bold;
-  margin-left: 100px;
-}
-
-@media screen and (max-width:400px) {
-  label,
-  input,
-  form button {
-    width: 100%;
-  }
-
-  label {
-    margin-bottom: 0.5rem;
-  }
-
-  input {
-    margin-bottom: 2rem;
-  }
-  form button {
-    margin: 0;
-  }
+form > div > label[data-v-6e1bd991] {
+  width: 9rem
 }
 </style>
 
-{% block login %}
-<form action={{ url_for('login') }} method=post>
-  {% if errors %}
-    <ul>
-    {% for error in errors %}
-      <li class=error>{{ error }}</li>
-    {% endfor %}
-    </ul>
-  {% endif %}
-  <h1>Log into my account</h1>
-  <div>
-    <label for=email>Email</label>
-    <input type=email id=email name=email required>
-  </div>
-  <div>
-    <label for=password>Password</label>
-    <input type=password id=password name=password required>
-  </div>
-  <button type=submit>Log in</button>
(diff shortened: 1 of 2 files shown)


templates/auth/login.html
```diff
@@ -8,78 +8,45 @@
...
-    margin-bottom: 0.5rem;
-  }
-
-  input {
-    margin-bottom: 2rem;
-  }
-  form button {
-    margin: 0;
-  }
+form > div > label[data-v-6e1bd991] {
```
vinyll, line 26: Nope, don't do that. The `data-v-6e1bd991` may change.

templates/auth/login.html
```diff
@@ -8,78 +8,45 @@
...
-  <div>
-    <label for=password>Password</label>
-    <input type=password id=password name=password required>
-  </div>
-  <button type=submit>Log in</button>
-  <p class=alternative><a href={{ url_for('signup') }}>I don't have an account yet!</a></p>
-</form>
+<section id=body>
+  {% block login %}
+  <x-form class=login action={{ url_for('login') }} method=post>
```
vinyll, line 33: for actual usage outside the demo you may want to say `<form is=x-form …>`.

templates/auth/login.html
```diff
@@ -8,78 +8,45 @@
...
-  }
-
-  input {
-    margin-bottom: 2rem;
-  }
-  form button {
-    margin: 0;
-  }
+form > div > label[data-v-6e1bd991] {
+  width: 9rem
```
vinyll, line 27: Why do you need this BTW?

templates/auth/login.html
```diff
@@ -8,78 +8,45 @@
...
 }
 </style>
 
-{% block login %}
-<form action={{ url_for('login') }} method=post>
-  {% if errors %}
-    <ul>
-    {% for error in errors %}
```
vinyll, line 67: You still need this, don't you?

templates/auth/login.html
```diff
@@ -8,78 +8,45 @@
...
-  <p class=alternative><a href={{ url_for('signup') }}>I don't have an account yet!</a></p>
-</form>
+<section id=body>
+  {% block login %}
+  <x-form class=login action={{ url_for('login') }} method=post>
+    <div class=form-row>
+      <label for=email>Email</label>
+      <input type=email name=email id=email>
+    </div>
+    <div class=form-row>
```
vinyll, line 38: `class=form-row` is actually not required: https://github.com/neomad-team/styleguide/blob/vuejs-neomad/src/components/x-form.vue#L19

We should adapt the styleguide demo also.

templates/auth/login.html
```diff
@@ -8,78 +8,45 @@
...
-  </div>
-  <button type=submit>Log in</button>
-  <p class=alternative><a href={{ url_for('signup') }}>I don't have an account yet!</a></p>
-</form>
+<section id=body>
+  {% block login %}
+  <x-form class=login action={{ url_for('login') }} method=post>
+    <div class=form-row>
+      <label for=email>Email</label>
+      <input type=email name=email id=email>
```
vinyll, line 36: `required`

templates/auth/login.html
```diff
@@ -8,78 +8,45 @@
...
+    </div>
+    <div class=form-row>
+      <label for=password>Password</label>
+      <input type=password id=password name=password required>
+    </div>
+    <input is=x-button class=primary type=submit name=login value=Login>
```
vinyll, line 42: Why does the text look so small on your screenshot

templates/auth/login.html
```diff
@@ -8,78 +8,45 @@
...
+    <hr>
+    <input type=submit name=signup value="Create an account" class=default>
+  </x-form>
+  {% endblock %}
+</section>
 {% endblock %}
+
+{% block extra_js %}
+<script src={{ url_for('static', filename='dist/build.js') }}></script>
```
vinyll, line 51: This shall probably end up in the layout.
We may also want to rename it to _dist/components.js_ or something (not sure neither about a _dist_ folder for Neomad.org of _js_ folder, or elsewhere).

vinyll: We shall probably also need a _styleguide.css_ or something.

vinyll: be aware that components call an external css! (how should we make this dependency??): https://github.com/neomad-team/styleguide/blob/vuejs-neomad/src/components/x-form.vue#L2
Not good 🤔

vinyll (changes requested):

vinyll:
I ❤️ed this result but I don't think we'll use it to go for css `@import`.
Should we close this PR @fuentesloic / @newick?

newick:
@vinyll if `@import`is the easiest way, i'm in

vinyll:
Closed for using native CSS instead.

PR #196, opened 2017-10-09
argument to filter list article by profile missing

fuentesloic:
Write by Python expert ;)

### DIFF
--- a/user/views.py
+++ b/user/views.py
@@ -13,8 +13,9 @@ from .models import User
 def profile(username):
     try:
         user = User.objects.get(slug=username)
-        articles = Article.objects(author=user)
-        if user != current_user:
+        if user == current_user:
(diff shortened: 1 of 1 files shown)


user/views.py
```diff
@@ -16,7 +15,7 @@ def profile(username):
         user = User.objects.get(slug=username)
         articles = Article.objects(author=user)
         if user != current_user:
-            articles = Article.published()
+            articles = Article.published(author=user)
```
vinyll:
```python
if user == current_user:
    artlcles = Article.objects(author=user)
else:
    articles = Article.published(author=user)
```

vinyll (changes requested):
Pretty good for a future python dev 😛

vinyll (approved):

PR #199, opened 2017-10-12
fixed error when empty current_location

vinyll:


fuentesloic (approved):

fuentesloic:
@vinyll review done! we are closeeeee

Issue #206, opened 2017-10-16
no title on avatar


fuentesloic:
We should add a title on an avatar in the home page

fuentesloic:
fix on #223

PR #223, opened 2017-10-29
add title on user img

fuentesloic:
fix the #206

### DIFF
--- a/templates/home.html
+++ b/templates/home.html
@@ -72,11 +72,11 @@
           <figure>
             <img 
               src={{ user.avatar }}
-              title={{ user.avatar }}
+              title={{ user.username }}
               width=100
               height=100
               onerror="this.src='{{ url_for('static', filename='images/avatar.png') }}'"
-              alt>
+              alt="photo of {{ user.username }}">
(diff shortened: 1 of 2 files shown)


templates/home.html
```diff
@@ -70,7 +70,13 @@
...
         <a href='{{ url_for_user(user) }}'>
           <figure>
-            <img src={{ user.avatar }} width=100 height=100 onerror="this.src='{{ url_for('static', filename='images/avatar.png') }}'" alt="">
+            <img 
+              src={{ user.avatar }}
+              title={{ user.avatar }}
+              width=100
+              height=100
+              onerror="this.src='{{ url_for('static', filename='images/avatar.png') }}'"
+              alt>
```
vinyll: Shouldn't we insert some alternative text here for accessibility purpose? Like "photo of {{ user }}"

vinyll (dismissed):

templates/home.html
fuentesloic: true and done :)

vinyll (approved):

PR #228, opened 2017-11-19
WIP - Quill editor

fuentesloic:


fuentesloic:
Ok I need help now I think with python or not...

- I removed the medium editor,
- I add Quill script/css,
- I move the editor JS from HTML to a real JS file,
- I create two editor, one for Title (without toolbox because it's just a title, no personalisation from user), second for content with toolbox (for now just simple style availables),
- I send through the form the Quill object according the good element on db

And ... I need you for this step...

Before you was catching the HTML content so that was easy to put it with python with `{{ article.content }}` but now it's an Objet wich need to be transform with Quill function call `setContents()` something like:
```
const editors = [{
  container: document.querySelector('#titleEditor'),
  content: {{ article.title }}
  }, {
  container: document.querySelector('#titleEditor'),
  content: {{ article.content }}
}]

editors.map(editor => editor.container.setContents(editor.content))
```

vinyll:
What does Python have to do with this? JS should do that job.
Python is only saving raw data, not specific objects.

Also why do we need that Quill for the title (even without toolbars)?

fuentesloic:
Oh I see so let's go for appendChild I was thinking that python could transform it

docker-compose.yml
```diff
@@ -19,4 +19,4 @@ services:
     volumes:
       - .:/app
     ports:
-      - 5000:80
+      - 5000:8080
```
vinyll: ❓

static/js/editor.js
```diff
@@ -0,0 +1,18 @@
+var editorTitle = new Quill('#editorTitle')
+
+var editorContent = new Quill('#editorContent', {
```
vinyll: Use `const` unless you have a good reason. And you certainly don't here ;)

static/js/editor.js
```diff
@@ -0,0 +1,18 @@
+var editorTitle = new Quill('#editorTitle')
```
vinyll: title don't need an editor

static/js/editor.js
```diff
@@ -0,0 +1,18 @@
...
+    toolbar: [
+      ['bold', 'italic'],
+      ['link', 'image']
+    ]
+  }
+})
+
+document.querySelector('form[edit]').addEventListener('submit', event => {
+  const form = event.target
+  form.title.value = editorTitle.getContents()
```
vinyll: Just get the title from the `textContent` [like we did](https://github.com/neomad-team/neomad.org/pull/228/files#diff-b2eb8af5867e7d92dac4de3e62798caeL105)

templates/blog/article.html
```diff
@@ -15,6 +15,7 @@
 
 {% block extra_css %}
 <link rel=stylesheet href={{ url_for('static', filename='css/article.css') }}>
+<link href=https://cdn.quilljs.com/1.3.4/quill.bubble.css rel="stylesheet">
```
vinyll, line 18: download it locally, and remove the quotes

templates/blog/article.html
```diff
@@ -45,10 +46,14 @@
   </div>
   {% endif %}
   
-  <h1 name=title>{{ article.title or 'Insert your title here' }}</h1>
-  <div class=content name=content>
+  <div id=editorTitle>
+    <h1>{{ article.title or 'Insert your title here' }}</h1>
```
vinyll: we need [this line](https://github.com/neomad-team/neomad.org/pull/228/files#diff-b2eb8af5867e7d92dac4de3e62798caeL48) to make the title editable

templates/blog/article.html
```diff
@@ -58,13 +63,8 @@ <h1 name=title>{{ article.title or 'Insert your title here' }}</h1>
...
 
 {% if edit %}
-<script src="//cdn.jsdelivr.net/medium-editor/latest/js/medium-editor.min.js"></script>
-<link rel=stylesheet href=//cdn.jsdelivr.net/medium-editor/latest/css/medium-editor.min.css type=text/css media=screen charset=utf-8>
-
 <link rel=stylesheet href=//netdna.bootstrapcdn.com/font-awesome/4.1.0/css/font-awesome.css>
-<link rel=stylesheet href=https://cdnjs.cloudflare.com/ajax/libs/medium-editor/5.23.0/css/themes/tim.min.css>
-<link rel=stylesheet href=https://cdn.rawgit.com/orthes/medium-editor-insert-plugin/3.0/dist/css/medium-editor-insert.min.css>
-<script src=https://cdn.rawgit.com/orthes/medium-editor-insert-plugin/3.0/dist/js/medium-editor-insert.min.js></script>
+<script src=https://cdn.quilljs.com/1.3.4/quill.js></script>
```
vinyll, line 64: download locally ;)

vinyll (changes requested):

static/css/article.css
```diff
@@ -37,16 +37,16 @@ article .date {
   font-size: .7em;
 }
 
-#editorTitle h1 {
+h1 {
```
vinyll: `article h1`?

static/js/editor.js
```diff
@@ -10,9 +9,11 @@ var editorContent = new Quill('#editorContent', {
   }
 })
 
+title.setAttribute('contenteditable', true)
+
 document.querySelector('form[edit]').addEventListener('submit', event => {
   const form = event.target
-  form.title.value = editorTitle.getContents()
-  form.content.value = editorContent.getContents()
+  form.title.value = title.innerHTML
```
vinyll: `form.title.value = title.textContent` I guess, as we don't want to allow HTML in the title. (we should even clean it BTW, but later I guess)

vinyll (commented):
Remove the word "WIP" if you expect a review ;)

templates/blog/article.html
```diff
@@ -45,10 +46,11 @@
   </div>
   {% endif %}
   
-  <h1 name=title>{{ article.title or 'Insert your title here' }}</h1>
-  <div class=content name=content>
+  <h1 name=title>{{ article.title or 'Insert your title here' }}</h1>
```
vinyll, line 49: remove the ending spaces

PR #230, opened 2018-01-23
Community mode - refactoring

vinyll:


vinyll:
Why so many commits? Let me cancel this and clarify…

PR #231, opened 2018-01-24
added names for new nomads list section on home

vinyll:


templates/home.html
```diff
@@ -71,13 +71,14 @@ <h2>Welcome to our last Neomads</h2>
...
           <figure>
-            <img 
+            <img
               src={{ user.avatar }}
               title={{ user.username }}
               width=100
               height=100
               onerror="this.src='{{ url_for('static', filename='images/avatar.png') }}'"
               alt="photo of {{ user.username }}">
+            <figcaption>{{ user }}</figcaption>
```
fuentesloic, line 81: Why `user`? should be `user.username` I think

fuentesloic (changes requested):

templates/home.html
vinyll: We say `{{ user }}` as we want the displayable version of user (ie: `user.__str__()`).
Python will render that, which is described by the _model_.
Therefore it will be `user.username` in most cases, but the model will be able to deal with it when the user does not have a username ;)

fuentesloic: ok I see

fuentesloic (approved):

PR #233, opened 2018-02-11
fix bugs on hero section then clean css

fuentesloic:


vinyll (approved):
the branch is called "arrow-down-position" and the Pr talk about hero ^^

PR #234, opened 2018-02-11
mobile nav

fuentesloic:


fuentesloic:
capture d ecran 2018-02-11 a 13 21 56-fullpage
capture d ecran 2018-02-11 a 13 22 00-fullpage

static/css/main.css
```diff
@@ -151,6 +150,67 @@ body > header nav .button:hover {
...
+    background-color: white;
+    padding: 3rem;
+    box-shadow: 1px 1px 2px #eee;
+  }
+  
+  body > header nav a {
+    font-size: 2.25rem;
+    line-height: 5rem;
+  }
+  #avatar-menu {
```
vinyll: ` #avatar-menu` vs `.profile-menu` is confusing

static/css/main.css
```diff
@@ -372,26 +432,17 @@ button:hover,
...
-@media screen and (max-width:700px) {
+@media screen and (max-width: 700px) {
   article {
     width: 100%;
     padding: 0 2rem;
   }
 }
 
-@media screen and (max-width:800px) {
+@media screen and (max-width: 800px) {
```
vinyll: Wondering if we should try to match with the mediaquery here above or if it's better to keep it on a different width.

templates/layout.html
```diff
@@ -17,7 +17,8 @@
 
 <header>
   <h1><a href=/>Neomad</a></h1>
-  <nav>
+  <a href=# class=mobile-nav>Menu</a>
```
vinyll: When inserting things like `<a href=#`, it's a sign it should be a `<button` ;)

vinyll (changes requested):
Moving from `a href=#` to `<button` will syntaxically more pertinent ;)

templates/layout.html
```diff
@@ -17,15 +17,15 @@
...
 <header>
   <h1><a href=/>Neomad</a></h1>
-  <a href=# class=mobile-nav>Menu</a>
+  <button class=mobile-nav>Menu</button>
   <nav class=mobile-close>
     <a href={{ url_for('article_list') }} title="Read Neomad's adventures">Read</a>
     <a href={{ url_for('around') }}>Around me</a>
     {% if current_user and current_user.is_authenticated %}
-    <div id=avatar-menu class=avatar>
+    <div id=avatar-nav class=avatar-nav>
```
vinyll: `id=avatar-nav class=avatar-nav`??
Reminder: you would use `class` if it's a reusable behavior; you would use `id` if the element is unique per page.
I think you did this to be pointed by the JS but it still doesn't make sense to me.

If we have only 1 per page, let's keep the id.
If we may use this behavior, let's keep the class and when refering to this specific item with a nested selector.

templates/layout.html
```diff
@@ -17,15 +17,15 @@
...
       <img src={{ current_user.avatar }} width=35 height=35 onerror="this.src='{{ url_for('static', filename='images/avatar.png') }}'" alt>
     </div>
-    <ul id=menu class=profile-menu>
+    <ul id=profile-menu class=profile-menu>
```
vinyll: Same here 🙃

static/css/main.css
```diff
@@ -82,10 +82,10 @@ body > header nav a {
   font-size: 1.5rem;
   text-decoration: none;
 }
-.avatar {
+.avatar-nav {
```
vinyll: `nav .avatar`?

static/css/main.css
```diff
@@ -432,14 +378,71 @@ button:hover,
...
+    font-size: 2.25rem;
+    line-height: 5rem;
+  }
+  
+  .avatar-nav {
+    display: none;
+    background: none;
+    border: none;
+  }
+  .profile-menu {
```
vinyll: `nav .menu`?

vinyll (changes requested):
I'm sure you have very good reasons to go for these kind of selectors but they don't appear obvious from the outside

static/css/main.css
```diff
@@ -372,26 +378,79 @@ button:hover,
...
+    right: 0;
+  }
+  nav .profile::before {
+    display: none;
+  }
+  nav .profile > li {
+    display: flex;
+    justify-content: center;
+  }
+  nav .profile > li > a{
```
vinyll: missing a space before bracket

static/css/main.css
```diff
@@ -372,26 +378,79 @@ button:hover,
...
+    padding: 3px 15px;
```
vinyll: Any specific reason to use `px` here vs `rem`?

fuentesloic: because nobody like css :)

static/css/main.css
fuentesloic: nianiania

vinyll (approved):

PR #235, opened 2018-02-11
Users on map

fuentesloic:
display all the users who authorize community to show them in your _around map_

### DIFF
--- a/around/views.py
+++ b/around/views.py
@@ -1,8 +1,10 @@
 from flask import render_template
 
 from core import app
+from user.models import User
 
 
 @app.route('/around/')
 def around():
-    return render_template('around/map.html')
+    users = User.objects.all()
+    return render_template('around/map.html', users=users)
--- a/around/views.py
+++ b/around/views.py
@@ -6,5 +6,5 @@ from user.models import User
 
 @app.route('/around/')
 def around():
-    users = User.objects.all()
+    users = User.objects(allow_community=True)
     return render_template('around/map.html', users=users)
--- a/static/js/around-map.js
+++ b/static/js/around-map.js
@@ -46,9 +46,29 @@ map.on('click', event => {
   }
 })
 
+// add user on map
+usersLocation.forEach(user => {
+  addUser(user)
+})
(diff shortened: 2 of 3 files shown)


around/views.py
```diff
@@ -1,8 +1,10 @@
 from flask import render_template
 
 from core import app
+from user.models import User
 
 
 @app.route('/around/')
 def around():
-    return render_template('around/map.html')
+    users = User.objects.all()
```
vinyll: You should only send users that permitted to do so here instead of filtering in the template:
`users = User.objects(allow_community=True)`

static/js/around-map.js
```diff
@@ -46,9 +46,29 @@ map.on('click', event => {
   }
 })
 
+// add user on map
+usersLocation.forEach(user => {
```
vinyll: `usersLocation.forEach(user => addUser(user))`

templates/around/map.html
```diff
@@ -71,7 +72,18 @@ <h2>Share your place</h2>
...
 {% block extra_js %}
   <script>
-    const userID = {{ current_user.id or 'null' }}
+    const usersLocation = []
+    {% for user in users %}
+      {% if not current_user and user.allow_community %}
+        usersLocation.push({ 
+          name: '{{ user.username }}',
+          position: {{ user.current_location }},
+          link: {{ url_for_user(user) }}
```
vinyll: `link: "{{ url_for_user(user) }}"` as it might contain special characters

templates/around/map.html
```diff
@@ -71,7 +72,18 @@ <h2>Share your place</h2>
...
+        })
+      {% endif %}
+    {% endfor %}
+
+    const userID = '{{ current_user.id or 'null' }}'
```
vinyll: I'm not sure what `const userID = 'null'` as as string would give?

vinyll (changes requested):

vinyll (approved):
Feels like an expert JS came throught here :p

PR #236, opened 2018-02-17
css fixs from dev test - fully responsive

fuentesloic:


static/css/main.css
```diff
@@ -243,7 +242,7 @@ figure {
 }
 .meta .reading-time::before {
   content: "-";
-  margin-right: 85rem;
+  margin-right: .85rem;
```
vinyll, line 245: ahahaha 😆

vinyll (approved):

static/css/main.css
fuentesloic: I needed 5 minutes to catch the error ^^

PR #238, opened 2018-02-18
User picture around me

fuentesloic:


templates/around/map.html
```diff
@@ -74,11 +74,12 @@ <h2>Share your place</h2>
   <script>
     const usersLocation = []
     {% for user in users %}
-      {% if not current_user %}
+      {% if user != current_user %}
         usersLocation.push({ 
           name: '{{ user.username }}',
```
vinyll, line 79: Makes me think that we should escape characters.
I the user enters something malicious it could screw…
Let's say his name is `john's doe`; the page breaks.
But if he's called `john';alert('xss attack')` then we're screwed.

static/css/around-map.css
```diff
@@ -335,6 +335,15 @@
   color: #297ddb;
 }
 
+.user-avatar {
```
vinyll: Why `.user-avatar` and not `.avatar`?
Do we have non user avatars?

fuentesloic: I would made a difference because `.avartar` is already use for the avatar in the header.

vinyll: I suppose they share a lot in common then?
For the specificities can't we have `header .avatar` and `.profile .avatar` or something?

fuentesloic: I already `header .avatar` so no conflict possible

fuentesloic:
capture d ecran 2018-02-18 a 16 38 10

templates/around/map.html
vinyll: We're actually [safe by default](http://flask.pocoo.org/docs/0.12/security/#cross-site-scripting-xss). 😅

static/css/around-map.css
vinyll: Ok, I see what you mean. It's basically the icon of the user on the map. Much different then.

vinyll (approved):

PR #247, opened 2018-03-15
added 'trips' to user menu

vinyll:


fuentesloic (approved):
why we didn't it before ... 👍

PR #254, opened 2018-03-23
Edit profile separate

vinyll:

PR #258, opened 2018-03-24
Article edit split

### LINKED ISSUE #244: Save and publish
Il semble que cette option ne fonctionne pas. Je ne trouve rien dans les publications

### LINKED ISSUE #256: Editing an article with an embeded media deletes the media
When inserting a youtube video, the content is replaced by an iFrame.
`https://www.youtube.com/watch?v=_lNoRfXY0dg` > `<iframe src=https://www.youtube.com/watch?v=_lNoRfXY0dg…`

When editing this article again, the iframe is removed and the url doesn't appear anymore.

Suggestion: When replacing the url with an iframe, keep the url after the video. It is also useful as a legend.

vinyll:
Replaces #254 
fix #256 
fix #244


templates/blog/article.html
```diff
@@ -4,14 +4,14 @@
...
-{% block extra_og %}
-{% if article.image %}
-<meta property=og:image content="{% if article.image[:4] == 'http' %}{{ article.image }}{% else %}{{ request.url_root[:-1] }}{{ article.image }}{% endif %}">
-{% endif %}
-<meta property=og:article:published_time content="{{ article.creation_date }}">
-<meta property=og:article:author content="{{ article.author|capitalize }}">
-<meta property=og:description content="Article « {{ article|striptags|truncate(200) }} », by {{ article.author|capitalize }} on {{ article.creation_date|datetime }}">
-{% endblock %}
+{% block extra_og -%}
+  {% if article.image -%}
```
fuentesloic, line 8: why you finish with `-%}`? new way to write ? should we use everywhere for `block` and `if` ?

templates/blog/article.html
```diff
@@ -31,84 +31,26 @@
...
-    form.published.value = event.explicitOriginalTarget.value
-  })
-</script>
-
-<script src={{ url_for('static', filename='js/medium-editor/video-button.js') }}></script>
-
-{% endif %}
+{% else %}
+  <aside>
+    <a href={{ url_for_user(article.author) }}>Read more from {{ article.author }}</a>
```
fuentesloic, line 53: I will add soon `button-action` class on this PR or after doesn't matter it's not mandatory

fuentesloic (approved):

templates/blog/article.html
vinyll: `-%}` is a writing I hadn't understood so far. It removes the next newline.
I guess we could use it everywhere we create a newline for the html

templates/blog/article.html
vinyll: why not `.button.action` instead?

PR #279, opened 2018-04-10
Set z-index offset according leaflet doc

### LINKED ISSUE #278: Around: users should be above the locations markers

fuentesloic:
fix #278 
http://leafletjs.com/reference-1.3.0.html#marker-zindexoffset

### DIFF
--- a/static/js/around-map.js
+++ b/static/js/around-map.js
@@ -69,6 +69,7 @@ function addUser (user) {
   })
 
   const userMarker = L.marker(user.position, {icon: userIcon})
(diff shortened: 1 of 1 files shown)


fuentesloic:
@vinyll ok for this one ?

vinyll (approved):
I had in mind to use 2 LayerGroups (1 for workspots, 1 for users), but let's go on z-index for now if it's doing the job!

fuentesloic:
Keep it in mind :)

PR #292, opened 2018-04-24
70 webmanifest

### LINKED ISSUE #70: Web App manifest for mobiles
We should be able to optimize the look and feel on a mobile device.
As a native app.

https://developer.mozilla.org/en-US/docs/Web/Manifest

Let's be lazy: https://app-manifest.firebaseapp.com/

fuentesloic:
fix #70

### DIFF
--- a/static/webmanifest.json
+++ b/static/webmanifest.json
@@ -4,7 +4,7 @@
   "description": "Find a local place to work, get connected to other remote workers.",
   "background_color": "#279e6a",
   "theme_color": "#279e6a",
-  "dir": "rtl",
+  "dir": "ltr",
   "lang": "en",
   "display": "standalone",
   "orientation": "portrait-primary"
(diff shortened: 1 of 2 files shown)


static/webmanifest.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "Neomad. The Open Digital-Nomad Community",
+  "short_name": "Neomad",
+  "description": "Find a local place to work, get connected to other remote workers.",
+  "background_color": "#279e6a",
+  "theme_color": "#279e6a",
+  "dir": "rtl",
```
vinyll: Are you sure about the right-to-left?

vinyll (changes requested):
Man, you're a machine 🏗

static/webmanifest.json
fuentesloic: Oh ....... I stopped coffee the morning ^^

vinyll (approved):

PR #297, opened 2018-04-29
remove read time

fuentesloic:


templates/blog/macros.html
```diff
@@ -10,7 +10,6 @@
       <div>
         <a href={{ url_for_user(article.author) }}>{{ article.author.username }}</a>
         <time datetime={{ article.creation_date }} pubdate>{{ article.creation_date.strftime('%d %B, %Y') }}</time>
-        <p class=reading-time><span>{{ (article.content|length / 6 / 250)|int }}</span> min read</p>
```
vinyll, line 13: could you try to do `article.content|striptags|length / 6 / 250)|int` and see what the reading time is with an image?

vinyll (commented):
I guess removing just the reading time from the template could do and fix that later.

However, [`striptags`](http://jinja.pocoo.org/docs/2.10/templates/#striptags) could actually do the job eventually…

PR #305, opened 2018-08-27
create new api for login from mobile app

fuentesloic:
I tried ... even not sure if I understand my own python ^^

### DIFF
--- a/api/views.py
+++ b/api/views.py
@@ -39,10 +39,13 @@ def api_user(id):
     return Response(json.dumps(data), mimetype='application/json')
 
 
-@app.route('/api/login/', methods=['get', 'post'])
+@app.route('/api/login/', methods=['post'])
 def api_login():
-    user = User.objects.get(email=request.form['email'])
+    try:
+        user = User.objects.get(email=request.form['email'])
+    except User.DoesNotExist:
+        return jsonify({'success': False}), 401
     if user.check_password(password=request.form['password']):
-        return jsonify(user)
+        return jsonify(user.to_dict())
     else:
-        return jsonify({'success': False})
+        return jsonify({'success': False}), 401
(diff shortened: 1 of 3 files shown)


api/views.py
```diff
@@ -1,10 +1,10 @@
 import json
 
 from flask import request, Response
```
vinyll: `from flask import request, Response, jsonify`

api/views.py
```diff
@@ -37,3 +37,12 @@ def api_user(id):
...
     return Response(json.dumps(data), mimetype='application/json')
+
+
+@app.route('/api/login/', methods=['get', 'post'])
+def api_login():
+    response = request.json
+    user = User.objects.get(email=request.form['email'])
+    if user.check_password(password=request.form['password']):
+        login_user(user)
+        return Response(json.dumps(user), mimetype='application/json')
```
vinyll:
```python
    return jsonify(user)
else:
    return jsonify({'success': False})
```

api/views.py
```diff
@@ -37,3 +37,12 @@ def api_user(id):
     if 'email' in data['socials']:
         data['socials']['email'] = 'hidden'
     return Response(json.dumps(data), mimetype='application/json')
+
+
+@app.route('/api/login/', methods=['get', 'post'])
+def api_login():
+    response = request.json
```
vinyll: seems like you're not using this

vinyll (changes requested):
Good job mate!

fuentesloic:
I'm more like a child using copy/paste than python dev 😛

fuentesloic:
@vinyll thx to have fixed python!!!

vinyll (dismissed):

vinyll (approved):
@fuentesloic Notice that when an error occurs, a status code of 400 or 401 is returned.
When status code is 200, it means it's successful and the response contains the JSON user object :)

fuentesloic:
got it!

PR #308, opened 2018-09-14
Set position api

fuentesloic:
Need API to set user postion from mobile App

### DIFF
--- a/api/views.py
+++ b/api/views.py
@@ -49,3 +49,22 @@ def api_login():
         return jsonify(user.to_dict())
     else:
         return jsonify({'success': False}), 401
+
+
+@app.route('/api/position/', methods=['post'])
+def api_position():
+    try:
+        user = User.objects.get(id=request.form['id']).to_mongo()
+    except User.DoesNotExist:
+        return jsonify({'success': False}), 401
+    if not user.locations:
+        user.locations = [UserLocation(position=request.form['position'])]
+        user.save()
+        return '', 201
+    else:
+        index = user.locations.count() - 1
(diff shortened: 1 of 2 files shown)


fuentesloic:
@vinyll as expected ... my test it's not working ^^ I let you fix it ? (and build real test 😆🚀 )

api/views.py
```diff
@@ -49,3 +49,22 @@ def api_login():
         return jsonify(user.to_dict())
     else:
         return jsonify({'success': False}), 401
+
+
+@app.route('/api/position/', methods=['post'])
+def api_position():
```
vinyll, line 55: Why not using https://github.com/neomad-team/neomad.org/blob/master/trips/views.py#L29 
?
We know it works and [it's tested already](https://github.com/neomad-team/neomad.org/blob/master/trips/test.py#L31)

fuentesloic: because we have a @login_required and how Neomad could know that user is logged (on mobile app) ?

vinyll: Mhh… we can't just send the point without login the user or anyone could spoof. #security-issue

We need to create an authentication token and send it in the headers (`Authentication` header).

vinyll: The best way would be to authenticate with Oauth2, but it may be overkill in our case.
Maybe we could just generate an temporary token when the user logs in from the mobile app and send the token when communicating with the API.
It's a weak authentication process, but at least it's one.

vinyll: Eventually we could just use the id for now for the sake of easiness, open a security issue and treat it in a second time.

PR #312, opened 2018-10-08
add date input to customize publish date value

fuentesloic:
following #311

### DIFF
--- a/blog/test.py
+++ b/blog/test.py
@@ -227,7 +227,8 @@ class ArticleTest(TestCase):
     def test_parse_steemit(self):
         url = 'https://steemit.com/travel/@silkroad40/a-set-of-pictures-from-turkmenistan-and-bukhara-uzbekistan-ein-paar-bilder-von-philipe-s-reise-durch-turkmenistan-und-bukhara'
         article = Article(title='Steemit article', content=f'<p>{url}</p>',
-                          publication_date=now, author=self.user).save()
+                          published=True, publication_date=now,
+                          author=self.user).save()
         response = self.client.get(f'/@{self.user.slug}/{article.slug}-'
                                    f'{article.id}/')
         self.assertEqual(response.status_code, 200)
--- a/blog/test.py
+++ b/blog/test.py
@@ -33,6 +33,7 @@ class ArticleTest(TestCase):
         Article(title='A title for article',
                 content='<p>content</p>',
                 publication_date=datetime.datetime.utcnow(),
+                published=True,
                 author=self.user).save()
         result = self.client.get('/articles/')
         self.assertIn(b'A title for article', result.data)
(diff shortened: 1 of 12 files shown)


blog/views.py
```diff
@@ -68,7 +68,7 @@ def article_edit(id):
         article.title = request.form.get('title')
         article.content = request.form.get('content')
         if bool(request.form.get('published')):
```
vinyll: Editing the publication date is probably not the best idea of the world.
The issues says we should, but the need is to preserve the initial publication date.
I would therefore only set a publication date when it has none yet.
Therefore, an article that has a publication date should never be able to change it; it's it initial publication date:

```python
if bool(request.form.get('published') and not  article.publication_date):
    article.publication_date = datetime.datetime.utcnow()
```

However, doing so, I'm not sure we could still unpublish an article 🤔

blog/views.py
```diff
@@ -67,10 +67,8 @@ def article_edit(id):
     if request.method == 'POST':
         article.title = request.form.get('title')
         article.content = request.form.get('content')
-        if bool(request.form.get('published')):
+        if bool(request.form.get('published') and not article.publication_date):
```
vinyll: We actually need to convert to boolean the `request.form.get('published')` part:

```python
if bool(request.form.get('published')) and not article.publication_date:
```

fuentesloic:
@vinyll Indeed doing this, we can't unpublished an article when it's have been published once.

Could we add an `booleanField` in article model to deal with ? not to hard after with article in production ?

blog/views.py
```diff
@@ -67,10 +69,8 @@ def article_edit(id):
     if request.method == 'POST':
         article.title = request.form.get('title')
         article.content = request.form.get('content')
-        if bool(request.form.get('published')):
+        if bool(request.form.get('published')) and not article.publication_date:
             article.publication_date = datetime.datetime.utcnow()
```
vinyll, line 74: If you create an `article.published` property, you need to use it for publishing / unpublishing:

```python
article.published = bool(request.form.get('published'))
if article.published and not article.publication_date:
    article.publication_date = datetime.datetime.utcnow()
```

vinyll (changes requested):

blog/views.py
fuentesloic: I have to add in the model right ?

vinyll: You're totally right my python friend. 🐍

fuentesloic:
@vinyll and now ??

blog/test.py
```diff
@@ -174,16 +175,26 @@ def test_language_detection(self):
...
 
     def test_unpublished_article_does_not_appear_in_articles(self):
         Article(title='title', content='<p>content</p>',
+                published=False,
                 publication_date=None,
                 author=self.user).save()
         result = self.client.get('/articles/')
         self.assertNotIn(b'<article class=preview>', result.data)
+        
+    def test_published_then_unpublished_article_does_not_appear_in_articles(self):
```
vinyll: the test does not publish then unpublish.
You may mean `test_unpublished_article_does_not_appear_in_articles` ?

vinyll (approved):

PR #325, opened 2020-01-10
adapt larger images for article reading

vinyll:


fuentesloic:
Not sure to understand the purpose or the problem that you tried to fixed ^^

vinyll:
Indeed.
The text is narrow, which is great for faster reading.
This ends to make pictures narrow too.

Screen Shot 2020-01-12 at 13 38 42


The script here permits the pictures to be centered and larger than the text.
If you find a better way (in CSS?) it's welcome.

fuentesloic:
I checked with medium articles (and other sources) and the images width fit the text width. Do you have an other examples?

vinyll:
This was intially a demand from Marion.
I saw this case with larger pics a few days ago but I can't find the example in my history

fuentesloic (approved):
IMO images should have the same width (excepted if it's an hero image at the top). But as it's a UI topic and because I'm definitivly a bad designer 🤣 maybe we could ask to our layer specialist Erick or our king designer Aymeric?
OR ... just try it and see if people agree with 😃😃😃
I will transform it into CSS rules in a close future.

PR #326, opened 2020-01-10
show user current location from leaflet map

vinyll:


static/js/utils.js
```diff
@@ -4,7 +4,7 @@ window.onload = _ => {
   if (menu && avatar ) avatar.addEventListener('click', _ => menu.classList.toggle('active'))
 }
 
-function alert (type, message, delay) {
+function notify (type, message, delay) {
```
fuentesloic, line 7: 👍

fuentesloic (approved):

fuentesloic (changes requested):
The icon in the button is not visible (on FF)
Capture d’écran, le 2020-01-11 à 08 43 54

fuentesloic (dismissed):

fuentesloic:
I made the change with 71ea4a8 to add location marker icon, I will add it in the styleguide export (so when our neomad repo is stable)
Capture d’écran, le 2020-01-11 à 09 35 40

fuentesloic (approved):

PR #332, opened 2020-01-18
add style for link in article

### LINKED ISSUE #242: Hyperlink are not visible on article
It seems that the hyperlink doesn't work in in last article (draft box).

fuentesloic:
fix #242

Link on article:
Capture d’écran, le 2020-01-18 à 12 42 35
Focus, active and hover effect: 
Capture d’écran, le 2020-01-18 à 12 42 40

### DIFF
--- a/static/css/article.css
+++ b/static/css/article.css
@@ -26,6 +26,15 @@ body {
   margin-top: 2rem;
 }
 
+.blog-article .content a {
+  color: var(--color-primary);
+  text-decoration: underline dotted var(--color-primary);
+}
+
+.blog-article .content a:hover {
+  text-decoration: underline;
+}
+
 .button-group {
   padding: 0 1rem;
   position: absolute;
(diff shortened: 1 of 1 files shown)


static/css/article.css
```diff
@@ -26,6 +26,17 @@ body {
   margin-top: 2rem;
 }
 
+.blog-article .content a {
+  color: var(--color-primary);
```
vinyll, line 30: Doesn't `color: inherit;` work?

static/css/article.css
```diff
@@ -26,6 +26,17 @@ body {
...
+  text-decoration: underline dotted var(--color-primary);
+}
+
+.blog-article .content a:active,
+.blog-article .content a:focus,
+.blog-article .content a:hover {
```
vinyll, line 36: Can't you just say `.blog-article .content a {`?

vinyll (approved):
with questions

static/css/article.css
fuentesloic: Nope. I guess it's because color is not defined for `a`.

static/css/article.css
fuentesloic: I could but why? should we not have effect because it's on an article??

PR #335, opened 2020-02-03
#224 : Fix error 500 when deleting a trip point

### LINKED ISSUE #224: error 500 when deleting a trip point
I deleted my Paris (thanks to the phone) point and the feedback I had was a 500 page

mlbiche:
Fix #224

### DIFF
--- a/user/views.py
+++ b/user/views.py
@@ -3,6 +3,7 @@ from datetime import datetime
 
 from flask import render_template, request, abort, redirect, url_for
 from flask_login import current_user, login_required
+from mongoengine.errors import DoesNotExist
 
 from core import app
 from core.helpers import url_for_user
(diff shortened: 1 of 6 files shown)


user/views.py
```diff
@@ -50,10 +51,12 @@ def privacy():
...
     user = User.objects.get(id=current_user.id)
-    user.locations.remove(user.locations.get(date=datetime.fromtimestamp(
-                                             date)))
-    user.save()
-    return redirect('privacy'), 204
+    try:
+        user.locations.remove(user.locations.get(date=datetime.fromtimestamp(date)))
+        user.save()
+        return redirect(url_for('privacy'))
+    except DoesNotExist:
```
vinyll, line 58: I think you can do `User.DoesNotExist` instead of importing `DoesNotExist` from mongoengine.errors

vinyll (approved):
with suggestion

user/views.py
mlbiche: I confirm that it is a `mongoengine.errors.DoesNotExist`that raised because the `get` is done on `user.locations` and not on `User`.

vinyll: What about `UserLocation.DoesNotExist`? Which is more explicit about the nature of the item not found.

mlbiche: I have just tried, the class `UserLocation`doesn't have a `DoesNotExist` attribute.

vinyll: Could be as it is an `EmbeddedDocument`…
Nevermind then, all good :+1:

PR #337, opened 2020-02-03
Update trip list to reverse chronological order

mlbiche:


user/views.py
```diff
@@ -43,7 +43,7 @@ def privacy():
         user = User.objects.get(id=current_user.id)
         return render_template('user/privacy.html',
                            user=user,
-                           locations=user.locations)
+                           locations=sorted(user.locations, key=lambda l: l.date, reverse=True))
```
vinyll: Locations are sorted by default on save.
I guess `locations=user.locations.reverse()` should be enough.

mlbiche: Oh sweet, way more efficient ! Thanks 🎉

user/views.py
```diff
@@ -43,7 +43,7 @@ def privacy():
         user = User.objects.get(id=current_user.id)
         return render_template('user/privacy.html',
                            user=user,
-                           locations=user.locations)
+                           locations=reversed(user.locations))
```
vinyll, line 46: Where does thie `reversed` function come from?

vinyll: Does this actually work?

mlbiche: This actually works ! After some research, I've found that the difference between `reversed` and `list.reverse` is that `reversed` returns the reversed list but leaves the original list unchanged while `list.reverse` reverses the list but does not return the result. To user `list.reverse`, I have to add the line `user.locations.reverse()` before the `return`. What do you prefer ?

More information about `reversed` : https://docs.python.org/fr/3/library/functions.html?highlight=reversed#reversed

vinyll: That's indeed simple and great!

vinyll (approved):

PR #338, opened 2020-02-03
Remove Docker installation

mlbiche:
While the docker folder was deleted, some configuration and the README were not updated according to the removed Docker installation.


readme.md
```diff
@@ -76,6 +58,10 @@ In the previous commands you created a file called _settings.py_ at the root of
 the project.
 Tune your settings there.
 
+Note that the project is dedicated the port 27018 to the Docker container running
+MongoDB. You can change it with the port you want with the envionment variable
+`DB_PORT` declared in **both** _install.py_ and _start.py_.
```
vinyll: Suggestion:
```
`DB_PORT` is an environment variable where your Mongo instance is running. The default for a native install (27017) is preserved. _Lazy installation_ runs on 27018 by default.
```

bin/install.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+export DB_PORT=27018 # Edit with the wanted port
```
vinyll: I'm not sure someone using the automagic installer will tune this; and I guess the variable is obvious enough to avoid the current comment.

bin/start.sh
```diff
@@ -1,5 +1,10 @@
 #!/bin/bash
 
+export DB_PORT=27018 # Edit with the wanted port
```
vinyll: Same here.
Eventually a comment that would invite to ensure that the same value is used than in `/bin/install.sh` could be helpful.

mlbiche:
@vinyll I have updated the code according to your feedbacks. You can review it again 👍

readme.md
```diff
@@ -76,6 +58,8 @@ In the previous commands you created a file called _settings.py_ at the root of
 the project.
 Tune your settings there.
 
+In _install.py_ and _start.py_, `DB_PORT` is an environment variable where your Mongo instance is running. The default for a native install (27017) is preserved. _Lazy installation_ runs on 27018 by default.
```
vinyll: > In _install.py_ and _start.py_, `DB_PORT` is an environment variable where your Mongo instance is running

`DB_PORT` is actually available for all, not limited to install and start scripts. I guess you can therefore get rid of "In _install.py_ and _start.py_, ".

mlbiche: Done !

vinyll (approved):

PR #343, opened 2020-02-05
Fix wrong port make fixtures

mlbiche:


readme.md
```diff
@@ -26,7 +26,7 @@ The next steps are also simple. Choose your story:
 - for lazy developers: go to [Lazy developer install](#lazy-developer-installation)
 - for developers who want to keep control: read the [Manual installation](#manual-installation)
 
-> You may then want to insert dummy data use `make fixtures`.
+> You may then want to insert dummy data use `DB_PORT={YOUR_PORT} make fixtures` (it is already done in th _Lazy installation_).
```
vinyll: What about moving this section to the `### Manual installation` section as it only concern this type of installation?
Also, "> You may then want to insert dummy data use `make fixtures`" should be enough. An advanced user should be able to make his way around without telling her how to tune her port (plus, she would eventually prefer to setup the settings.py)

vinyll (approved):

PR #345, opened 2020-02-06
#334 : Fix the wrong error codes on error pages

### LINKED ISSUE #334: Fix error page with wrong error code
When the API raise an error, the error page _error.html_ is rendered with the error code displayed. However, if we check the full response received : indeed, the data contains the _error.html_ content, but the response code is 200 and not the expected error code.

An example of error page with the wrong error code

mlbiche:
Fix #334

### DIFF
--- a/core/views.py
+++ b/core/views.py
@@ -17,12 +17,12 @@ def home():
 
 @app.errorhandler(NotFound)
 def error_404(e):
-    return render_template('error.html', code=404)
+    return render_template('error.html', code=404), 404
 
 @app.errorhandler(Gone)
 def error_410(e):
-    return render_template('error.html', code=410)
+    return render_template('error.html', code=410), 410
 
 @app.errorhandler(InternalServerError)
 def error_500(e):
-    return render_template('error.html', code=500)
+    return render_template('error.html', code=500), 500


fuentesloic:
LGTM but I'm not the good flask reviewer 😛 
@vinyll up?

core/views.py
```diff
@@ -17,12 +17,12 @@ def home():
 
 @app.errorhandler(NotFound)
 def error_404(e):
-    return render_template('error.html', code=404)
+    return render_template('error.html', code=404), 404
```
vinyll, line 20: Why do we need both in render_template and in return?

mlbiche: It is done this way in [Flask documentation](https://flask.palletsprojects.com/en/1.1.x/quickstart/#rendering-templates). And I think it also allows to return the error code I have added and not only the template.

vinyll: the flask documentation indeed mentions that [it's a parameter of return](https://flask.palletsprojects.com/en/1.1.x/quickstart/#redirects-and-errors).
`code` therefore has no effect in `render_template` and should be removed to avoid confusion.

mlbiche: The parameter `code` is used in _error.html_ to display the error code in the view (line 13).

vinyll (approved):

PR #346, opened 2020-03-24
Refactor localization

### LINKED ISSUE #324: Geolocation is prompted on every page
Geolocation is essential when:
-  first loading a page for updating the user location
- looking at what is around me
- adding a new location

When I just provided my location and I refresh or browse the pages I shouldn't be prompted anymore.

Maybe just detecting a cookie with an expiration would do the job?

### LINKED ISSUE #349: Focus "Around you" map on user location
When opening "Around you" map, it show the worldwide map while what the user is looking for is POI around him/her.

mlbiche:
- Fix localization asked on each pages (Fix #324 )
- Clarify localization code by centralising its code

EDIT :
- Fix Around map not focused on user context (Fix #349 )

### DIFF
--- a/static/js/map-around.js
+++ b/static/js/map-around.js
@@ -1,6 +1,7 @@
 // POIS creation
 const worker = new Worker('/static/js/webworker-around.js')
 let pois = []
+let currentUserLocation
 
 function addPoi (poi) {
   const marker = L.marker(poi.location, {icon: markerIcon, alt: poi.name})
@@ -43,6 +44,11 @@ usersLocation.filter(user => user.position.length).forEach(user => addUser(user)
 window.onhashchange = _ => {
   const hash = getHash()
   hashMarker(hash)
+}
+
+function userLocalized(userLocation) {
+  currentUserLocation = userLocation
+  moveTo(currentUserLocation)
 }
 
 // functions
--- a/static/js/map-around.js
+++ b/static/js/map-around.js
@@ -1,7 +1,6 @@
 // POIS creation
 const worker = new Worker('/static/js/webworker-around.js')
 let pois = []
-let currentUserLocation
 
 function addPoi (poi) {
   const marker = L.marker(poi.location, {icon: markerIcon, alt: poi.name})
@@ -47,8 +46,7 @@ window.onhashchange = _ => {
 }
 
 function userLocalized(userLocation) {
-  currentUserLocation = userLocation
-  moveTo(currentUserLocation)
+  moveTo(userLocation)
 }
 
 // functions
--- a/static/js/utils.js
+++ b/static/js/utils.js
@@ -22,24 +22,3 @@ function notify (type, message, delay) {
     }, delay || 5000)
   }
 }
-
-function getPosition() {
-  return new Promise(resolve => {
-    const fail = _ => {
-      notify('error', 'You geolocalisation went wrong, please try again.')
-      resolve([])
-    }
-
-    const success = position => {
-      const { latitude, longitude } = position.coords
-      resolve([latitude, longitude])
-    }
-
-    const options = {
-      enableHighAccuracy: true,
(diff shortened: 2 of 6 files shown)


fuentesloic:
Just a question relative to flask I guess. Does it make sens to have big template because we set all the `js` on it? is not better to split html/js?

mlbiche:
@fuentesloic Not sure about what you meant through your question. 🤔 Do you have an example ? Are you talking about the localize.html file ? I am pretty new to Flask and Jinja so, I am not trying to change the existing architecture, just to refactor some pieces of code.

fuentesloic:
@vinyll up?

static/js/utils.js
```diff
@@ -21,25 +21,4 @@ function notify (type, message, delay) {
...
-
-    const options = {
-      enableHighAccuracy: true,
-      timeout: 5000,
-      maximumAge: 0
-    }
-    navigator.geolocation.getCurrentPosition(success, fail, options)
-  })
-}
+}
```
vinyll: Missing an newline

templates/script/localize.html
```diff
@@ -1,13 +1,31 @@
...
+          })
+
+          localizeUserPosition = [currentPosition.coords.latitude, currentPosition.coords.longitude]
+          sessionStorage.setItem('currentPosition', JSON.stringify(localizeUserPosition))
+
+          if (window.location.pathname === "{{ url_for_trips(current_user) }}" && res.status === 201)
+            window.location.reload()
         })
-      })
+      }
```
vinyll: I see this code was moved from utils.js to an includable HTML. It is solving something?
The benefit with this code is utils.js is that it it cached by the browser for all HTML pages.
the _templates/script/localize.html_ is not being cached as it is get included _as is_ into each HTML page.

static/js/utils.js
```diff
@@ -1,10 +1,38 @@
...
+        window.location.reload()
+    })
+  }
+  localizeUserPosition = JSON.parse(storedPositionJSON)
+
+  if (window.location.pathname === "{{ url_for('around') }}")
+    moveTo(localizeUserPosition)
+  if (window.location.pathname === "{{ url_for('form_spot') }}")
+    userPosition = localizeUserPosition
+}
```
vinyll, line 34: I guess this means that we want a specific action once the usser is localized.
It could be translated with something like:
```
if(userLocalized) userLocalized(localizeUserPosition)
```

Where `userLocalized` would be a function that is specific within the templates "around" and "form_spot"

templates/around/form_spot.html
```diff
@@ -51,9 +51,7 @@ <h2>Share your place</h2>
 
 {% block extra_js %}
   <script>
-    let userLocation = {{ current_user.current_location }}
-    // try to get browser position from promise in utils if the user do not share its position
-    if (!userLocation) userLocation = getPosition()
+    let userPosition // NOTE : Its value is affected from localize.html script
```
vinyll: Not sure if this is actually useful; `let` is local scoping.

templates/script/location-trips.html
```diff
@@ -1,5 +1,4 @@
 <script>
-	var current_location = {{ user.current_location or [] }}
 	var locations = [
```
vinyll, line 2: Not sure why we have 4 spaces in this file while it's 2 spaces all around.
I not this is out of this PR, just noticing.

mlbiche:
Also fix #349

static/js/map-around.js
```diff
@@ -1,6 +1,7 @@
 // POIS creation
 const worker = new Worker('/static/js/webworker-around.js')
 let pois = []
+let currentUserLocation
```
vinyll: Why do we need this here as `currentUserLocation` is only used in the `userLocalized` function?

vinyll (approved):
:)

PR #348, opened 2020-05-09
Systemd service

vinyll:
Permits to start and reload the server.
Hence to deploy with more stability.

### DIFF
--- a/neomad.service
+++ b/neomad.service
@@ -12,8 +12,6 @@ RestartSec=3
 StandardOutput=syslog
 StandardError=syslog
 SyslogIdentifier=neomad
-#User=<alternate user>
-#Group=<alternate group>
 Environment=PORT=8080
 
 [Install]
--- a/neomad.service
+++ b/neomad.service
@@ -1,0 +1,20 @@
+[Unit]
+Description=Neomad.org website
+After=network.target
+
+[Service]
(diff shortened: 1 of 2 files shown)


mlbiche:
Ok for me. Maybe the copy of the service to the systemd folder is missing somewhere ?

mlbiche (dismissed):
Ok for me. Maybe the copy of the service to the systemd folder is missing somewhere ?

neomad.service
```diff
@@ -0,0 +1,20 @@
...
+ExecStart=/home/ubuntu/prod/venv/bin/gunicorn app -w 3 --timeout 60
+# Required on some systems
+WorkingDirectory=/home/ubuntu/prod
+Restart=always
+RestartSec=3
+# Output to syslog
+StandardOutput=syslog
+StandardError=syslog
+SyslogIdentifier=neomad
+#User=<alternate user>
```
fuentesloic: As it's commented, why not removing it? same below

fuentesloic (approved):

mlbiche (approved):
LGTM, still the same comment about the service copy to systemd folder ?

vinyll:
> LGTM, still the same comment about the service copy to systemd folder ?

It's a symlink rather than a copy here.
The whole process for installation in production is not set in the makefile, it's hand made once with sudo, so it would be a different mission.

